PR #11, opened 2019-07-02
client beat interval should be customized at instance level.

IanCao:


IanCao:
#7

IanCao:
> the param semaphore of the function sendInstanceBeat is not necessary, we can get from BeatReactor, like br.semaphore

yes i‘ll fix

IanCao:
> the param semaphore of the function sendInstanceBeat is not necessary, we can get from BeatReactor, like br.semaphore

i have updated

PR #44, opened 2020-03-10
Feature auth sys

chuntaojun:
Add the new permission control function of nacos 1.2.0


common/nacos_server/nacos_server.go
```diff
@@ -32,18 +34,29 @@ type NacosServer struct {
 	vipSrvRefInterMills int64
 }
 
-func NewNacosServer(serverList []constant.ServerConfig, httpAgent http_agent.IHttpAgent, timeoutMs uint64, endpoint string) (NacosServer, error) {
+func NewNacosServer(serverList []constant.ServerConfig, clientCfg constant.ClientConfig, httpAgent http_agent.IHttpAgent, timeoutMs uint64, endpoint string) (NacosServer, error) {
 	if len(serverList) == 0 && endpoint == "" {
 		return NacosServer{}, errors.New("both serverlist  and  endpoint are empty")
 	}
+
+	securityLogin := security.NewAuthClient(clientCfg, serverList, httpAgent)
```
lzp0412, line 42: I think it's not need to create authClient when username or password is empty

chuntaojun: > I think it's not need to create authClient when username or password is empty

ok，I'll solve it

chuntaojun: > I think it's not need to create authClient when username or password is empty

I don't think so, but I think it's fair to say that inside the AuthClient it's just a matter of making a decision about whether or not the username is empty or whether or not you want to do the logic, because otherwise you're going to have to make a nil decision wherever you might be using AuthClient

PR #121, opened 2020-09-28
encode params before callConfigServer

### LINKED ISSUE #119: ConfigParam.Content 字符串中带有 & 无法推送到 nacos
在 PublishConfig 的时候，字符串中带有 & 的怎么处理呢？& 字符串后面的内容全部被丢弃了

mark4z:
Fixes #119

关于为什么没有Decode的部分：

Nacos server configController会解码params， 所以拉取配置时不必解码

### DIFF
--- a/common/http_agent/http_agent.go
+++ b/common/http_agent/http_agent.go
@@ -71,6 +71,8 @@ func (agent *HttpAgent) RequestOnlyResult(method string, path string, header htt
 }
 
 func (agent *HttpAgent) Request(method string, path string, header http.Header, timeoutMs uint64, params map[string]string) (response *http.Response, err error) {
+	util.EncodingParams(params)
+
 	switch method {
 	case http.MethodGet:
 		response, err = agent.Get(path, header, timeoutMs, params)
--- a/common/http_agent/http_agent.go
+++ b/common/http_agent/http_agent.go
@@ -92,6 +92,7 @@ func (agent *HttpAgent) Request(method string, path string, header http.Header, 
 }
 func (agent *HttpAgent) Post(path string, header http.Header, timeoutMs uint64,
 	params map[string]string) (response *http.Response, err error) {
+	util.EncodingParams(params)
 	return post(path, header, timeoutMs, params)
 }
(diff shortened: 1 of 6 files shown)


util/http.go
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package util
+
+import "net/url"
+
+func EncodingParams(params map[string]string) map[string]string {
```
sanxun0325, line 21: maybe it's better way.
```
 func EncodingParams(params map[string]string) {
	for param, value := range params {
		params [param] = url.QueryEscape(value)
 	}
 }
```

sanxun0325: EncodingParams func should be used in http_agent#Request func

mark4z: You are right, it's a mistake,  maybe java has deeply affected me. 
And now, it run  expected

sanxun0325 (approved):

common/http_agent/http_agent.go
```diff
@@ -71,6 +71,8 @@ func (agent *HttpAgent) RequestOnlyResult(method string, path string, header htt
 }
 
 func (agent *HttpAgent) Request(method string, path string, header http.Header, timeoutMs uint64, params map[string]string) (response *http.Response, err error) {
+	util.EncodingParams(params)
```
sanxun0325, line 74: I think maybe just encode it in a POST request.More testing can be done.

sanxun0325 (changes requested):

common/http_agent/http_agent.go
mark4z: OK，how about PUT method?

mark4z: and when I add test method in client_test, the test will failed with no logs, what I should do?

sanxun0325: dingding:wg46mjc，钉钉聊一下吧

PR #146, opened 2020-11-14
logger add rwmutex

Penglq:
My english is poor.
多次实例化client时，会多次执行InitLogger去给logger赋值，此时有协程异步打日志会读取logger，出现数据竞争

### DIFF
--- a/clients/naming_client/host_reator.go
+++ b/clients/naming_client/host_reator.go
@@ -54,7 +54,7 @@ func NewHostReactor(serviceProxy NamingProxy, cacheDir string, updateThreadNum i
 		updateTimeMap:        cache.NewConcurrentMap(),
 		updateCacheWhenEmpty: updateCacheWhenEmpty,
 	}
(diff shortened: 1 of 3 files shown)


common/logger/logger.go
```diff
@@ -30,6 +31,7 @@ import (
 
 var (
 	logger Logger
+	loggerRWMutex = sync.RWMutex{}
```
sanxun0325, line 34: In most cases of using nacos-sdk-go, users do not initialize multiple nacos clients in the program, and most users choose to override the logger of nacos-sdk-go.

sanxun0325: Currently, locks are not necessary in this PR, there is no guarantee that each client will use its own logger.

sanxun0325: Perhaps we could consider using Sync.once. To ensure that the client.CreateNamingClient() is executed successfully only Once.

PR #147, opened 2020-11-16
logger add sync.once

Penglq:
Increase the lock when setting log

### DIFF
--- a/clients/naming_client/host_reator.go
+++ b/clients/naming_client/host_reator.go
@@ -54,7 +54,7 @@ func NewHostReactor(serviceProxy NamingProxy, cacheDir string, updateThreadNum i
 		updateTimeMap:        cache.NewConcurrentMap(),
 		updateCacheWhenEmpty: updateCacheWhenEmpty,
 	}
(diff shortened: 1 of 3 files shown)


common/logger/logger.go
```diff
@@ -83,15 +85,17 @@ func init() {
...
-	logLevel := getLogLevel(config.Level)
-	encoder := getEncoder()
-	writer, err := getWriter(config.OutputPath, config.RotationTime, config.MaxAge)
-	if err != nil {
-		return
-	}
-	core := zapcore.NewCore(zapcore.NewConsoleEncoder(encoder), zapcore.AddSync(writer), logLevel)
-	zaplogger := zap.New(core, zap.AddCallerSkip(1))
-	logger = &NacosLogger{zaplogger.Sugar()}
+	loggerOnce.Do(func() {
```
sanxun0325, line 88: More talk is needed about loggers, and controlling logger Settings through Sync.once() may not be the best approach, as it can confuse users with logging when setting up multiple clients.

sanxun0325 (changes requested):

PR #153, opened 2020-11-26
filter loopback ip

Penglq:
ip
The LocalIP() method will get lo:1's IP address


sanxun0325:
Thank you for contribution.

Issue #166, opened 2020-12-09
ListenConfig  read config from file log err


christophe-liwei:
```
content, fileErr := cache.ReadConfigFromFile(key, client.configCacheDir)
if fileErr != nil {
logger.Errorf("[cache.ReadConfigFromFile] error: %+v", err)
}
```
这里的日志打印不对

sanxun0325:
@christophe-liwei Thanks for the reminder. It will be modified later.If you are interested, welcome to submit PR.

RayWong-codingPoet:
I'll try to fix this issue.

PR #168, opened 2020-12-11
Fix the issue of incorrect err display in ListenConfig.

RayWong-codingPoet:
FIX: https://github.com/nacos-group/nacos-sdk-go/issues/166


sanxun0325 (commented):
Thanks for your contribution ，look forward to more

sanxun0325 (commented):
Thanks for your contribution ，look forward to more

PR #193, opened 2021-03-11
Fix default timeoutms config

SpecialYang:
When users don't set timeoutms config, we should set the default value according the comment of this config.


sanxun0325 (approved):
Thank you for your contribution

PR #196, opened 2021-03-24
Trigger callback when subscribe service

lzp0412:
Currently, when the service information is cached in the local file, the callback function will not be triggered when we call the subscribe method. So we should trigger a callback when subscribe is called.

### DIFF
--- a/clients/naming_client/naming_client.go
+++ b/clients/naming_client/naming_client.go
@@ -297,10 +297,11 @@ func (sc *NamingClient) Subscribe(param *vo.SubscribeParam) error {
 	}
 
 	sc.subCallback.AddCallbackFuncs(util.GetGroupName(param.ServiceName, param.GroupName), strings.Join(param.Clusters, ","), &param.SubscribeCallback)
-	_, err := sc.GetService(serviceParam)
+	svc, err := sc.GetService(serviceParam)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


zhaoyunxing92 (approved):
agree

clients/naming_client/naming_client.go
```diff
@@ -297,10 +297,11 @@ func (sc *NamingClient) Subscribe(param *vo.SubscribeParam) error {
 	}
 
 	sc.subCallback.AddCallbackFuncs(util.GetGroupName(param.ServiceName, param.GroupName), strings.Join(param.Clusters, ","), &param.SubscribeCallback)
-	_, err := sc.GetService(serviceParam)
+	svc, err := sc.GetService(serviceParam)
 	if err != nil {
 		return err
 	}
+	sc.subCallback.ServiceChanged(&svc)
```
sanxun0325, line 304: 这样改后：当不使用本地缓存时（NotLoadCacheAtStart=true），SubscribeCallback func将被执行两次。
可以加入判断：
> if !sc.hostReactor.serviceProxy.clientConfig.NotLoadCacheAtStart {
> 		sc.subCallback.ServiceChanged(&svc)
> 	}

lzp0412: DONE

PR #212, opened 2021-04-15
[BUG 209] remove the return error when not find the config

### LINKED ISSUE #209: The config does not exist shouldn't make as an error.
The code:
https://github.com/nacos-group/nacos-sdk-go/blob/c9952e6dde21b58812a72c02cd79cb243a02f7a7/clients/config_client/config_client.go#L200

excepted:
only catch the 403 error like Java client
image

brothelul:
remove the return error when not find the config.
Fixes #209


sanxun0325 (approved):
LGTM

PR #213, opened 2021-04-16
fix nil pointer exception bug

DMwangnima:
When connection has not been established, close it directly would cause nil pointer exception.


sanxun0325 (approved):
LGTM

Issue #239, opened 2021-04-30
a panic i found


lppklm:
hello, 

there is a panic!

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb9ed8f]

goroutine 10626879 [running]:
api/vendor/github.com/nacos-group/nacos-sdk-go/clients/naming_client.(*PushReceiver).startServer.func1(0x0, 0xc003e379e0)
    src/api/vendor/github.com/nacos-group/nacos-sdk-go/clients/naming_client/push_receiver.go:94 +0x2f
created by api/vendor/github.com/nacos-group/nacos-sdk-go/clients/naming_client.(*PushReceiver).startServer
    src/api/vendor/github.com/nacos-group/nacos-sdk-go/clients/naming_client/push_receiver.go:93 +0x27d

i found a question in this function below:

```
func (us *PushReceiver) startServer() {
    var conn *net.UDPConn

    for i := 0; i < 3; i++ {
        r := rand.New(rand.NewSource(time.Now().UnixNano()))
        port := r.Intn(1000) + 54951
        us.port = port
        conn1, ok := us.tryListen()

        if ok {
            conn = conn1
            logger.Infof("udp server start, port: " + strconv.Itoa(port))
            break
        }

        if !ok && i == 2 {
            logger.Errorf("failed to start udp server after trying 3 times.")
        }
    }

    go func() {
        defer conn.Close()
        for {
            us.handleClient(conn)
        }
    }()
}
```

after retry 3 times,if the "conn" isn`t  assign a value, the panic will happened immediately~

i think the sdk can make it more friendly,  and you? thanks!

sanxun0325:
Yes, it can be optimized here.

after retry 3 times,if the "conn" isn`t assign a value,conn will be judged in handClient func , if it is nil then TryListen

himsngh:
Hey I would like to work on this issue.

sanxun0325:
@himsngh Looking forward to your PR

himsngh:
@sanxun0325 I did handle the panic as you mentioned although I have one minor doubt as to why we are doing it in the handle client instead we should have returned the function from the start-server only if it is nil even after the 3 iterations.
PS: I am new to all this so please do bear with me and do let me know if there is something else to this I will make the changes.

sanxun0325:
[...]

PR #243, opened 2021-05-02
handle panic push_receiver

himsngh:
issue #239


clients/naming_client/push_receiver.go
```diff
@@ -103,6 +103,16 @@ func (us *PushReceiver) startServer() {
 }
 
 func (us *PushReceiver) handleClient(conn *net.UDPConn) {
+
+	if conn == nil {
+		ok := false
+		conn, ok = us.tryListen()
```
sanxun0325, line 109: tryListen should be extract from StartServer() func. because us.tryListen () does not change the port.In Handleclient, you also need to interval the fetch of conn, for example, 5S once.Otherwise, the outer for loop will consume more resources.

应该将startServer func中获取tryListen 提取func, 目前这样直接tryListen 并没有换端口操作。 在handleClient中还需要将获取conn的操作进行间隔，例如5S一次。否则外层for循环会消耗较多资源。

sanxun0325 (changes requested):

clients/naming_client/push_receiver.go
himsngh: Okay yes, I got your point. One thing what do you mean when you say  "you also need to interval the fetch of conn, for example, 5S once" .

sanxun0325:
@himsngh 我重构了一下pr，有空可以review下

sanxun0325 (approved):

Issue #250, opened 2021-05-24
Nacos should not consume resources simply by being imported as a dependancy.


wcs1only:
We're looking to add nacos support as an optional feature. We attempted to add this support to an upcoming release, however, we noticed a significant CPU usage regression simply from importing the nacos-sdk-go package, whether or not our users had enabled the optional feature or not.

![Screen Shot 2021-05-21 at 1 34 05 PM](https://user-images.githubusercontent.com/7329298/119406813-29f67280-bc98-11eb-8126-bfcb0c116304.png)

After some profiling, we discovered that this extra CPU cost was caused by a goroutine created during init() of this module.

![Screen Shot 2021-05-24 at 2 01 38 PM](https://user-images.githubusercontent.com/7329298/119407144-a8ebab00-bc98-11eb-8a4a-23cd6ce286aa.png)

The goroutine in question is created here:
https://github.com/nacos-group/nacos-sdk-go/blob/master/clients/config_client/config_client.go#L81

Proposed fixes:
1. Can this delayScheduler goroutine be created with https://golang.org/pkg/sync/#Once the first time it is needed, rather than during module init() ?
2. It would seem like there may be some lock contention on the concurrentMap this goroutine is using. Perhaps its timer event is too agressive?

sanxun0325:
@wcs1only Yes, there is something wrong here, and I will fix it.

PR #251, opened 2021-05-26
remove init to start the scheduling task

sanxun0325:
fix：https://github.com/nacos-group/nacos-sdk-go/issues/250

The scheduling task is started when the Config Client is created.


lzp0412 (approved):
LGTM

Issue #262, opened 2021-06-09
Random ports have potential risks of port conflicts with other service


JoshuaAndrew:
https://github.com/nacos-group/nacos-sdk-go/blob/master/clients/naming_client/push_receiver.go#L77

#### What happened:
  push receiver port conflicts with other service
#### What you expected to happen:
  user can configure push receiver port by environment variable, If there is no environment variable, the default port is used

JoshuaAndrew:
@sanxun0325  @lzp0412  cc

sanxun0325:
I think if there is a conflict, it retries to get a new port

PR #263, opened 2021-06-10
通过环境变量指定 push receiver 监听端口重试次数失败是否退出

tedli:
#262 

case 1:
```
PUSH_RECEIVER_LISTEN_PORT=12345

-> port = 12345
```

case 2:
```
PUSH_RECEIVER_LISTEN_PORT=12345~23456

-> port = rand(12345~23456)
```

case 3:
```
PUSH_RECEIVER_RETRY_TIMES=5

-> for i := 0; i < 5; i++ {
	tryListen()
}
```

case 4:
```
PUSH_RECEIVER_EXIT_CODE_AFTER_RETRIED=8

-> for i := 0; i < 5; i++ {
	tryListen()

	if i == 4 {
		os.Exit(8)
	}
}
```

### DIFF
--- a/clients/naming_client/push_receiver.go
+++ b/clients/naming_client/push_receiver.go
@@ -23,6 +23,8 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net"
+	"os"
+	"regexp"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


clients/naming_client/push_receiver.go
```diff
@@ -70,22 +72,112 @@ func (us *PushReceiver) tryListen() (*net.UDPConn, bool) {
...
 
 		if ok {
 			conn = conn1
-			logger.Infof("udp server start, port: " + strconv.Itoa(port))
+			logger.Infof("udp server start, port: %d", us.port)
 			return conn
 		}
 
-		if !ok && i == 2 {
+		if !ok && i == retryTimes-1 {
```
rickey17, line 176: use enable env config to control os exit?

rickey17 (approved):

sanxun0325 (changes requested):
@tedli When the net.ListenUDP func() listens on the same port, it throws an exception and then refetches the port. this pr doesn't seem necessary

PR #288, opened 2021-07-28
Use SetIfAbsent instead of Set to avoid dup sendInstanceBeat task

YonkaFang:


clients/naming_client/beat_reactor.go
```diff
@@ -61,8 +61,9 @@ func buildKey(serviceName string, ip string, port uint64) string {
 func (br *BeatReactor) AddBeatInfo(serviceName string, beatInfo model.BeatInfo) {
 	logger.Infof("adding beat: <%s> to beat map", util.ToJsonString(beatInfo))
 	k := buildKey(serviceName, beatInfo.Ip, beatInfo.Port)
-	br.beatMap.Set(k, &beatInfo)
-	go br.sendInstanceBeat(k, &beatInfo)
+	if br.beatMap.SetIfAbsent(k, &beatInfo) {
```
sanxun0325, line 64: use reflect.DeepEqual contrast beatInfo

sanxun0325: @YonkaFang is there any progress?

YonkaFang: I still think it'not proper to check the beatInfo-change thing here. As when beatInfo actually changes, override it without stop the relevant task will bring a goroutine leak.

YonkaFang: To complete this case, it depends on the upper layer to do a check-and-then or stop-then-add action.

sanxun0325: Added beatInfo comparison logic

1. If the beatInfo is exactly the same, nothing needs to be done
2. If no, check whether the current beatInfo exists in beatMap
             if exists, then RemoveBeatInfo func is executed

sanxun0325: What do you think about this?
```
func (br *BeatReactor) AddBeatInfo(serviceName string, beatInfo model.BeatInfo) {
	logger.Infof("adding beat: <%s> to beat map", util.ToJsonString(beatInfo))
	k := buildKey(serviceName, beatInfo.Ip, beatInfo.Port)
	if _, ok := br.beatMap.Get(k); ok {
		br.RemoveBeatInfo(serviceName, beatInfo.Ip, beatInfo.Port)
	}
	br.beatMap.Set(k, &beatInfo)
	go br.sendInstanceBeat(k, &beatInfo)
}
```

YonkaFang: Note that the get-then-remove-then-set action is not atomic. It still may cause some dup tasks/goroutines.
A sync map does not ensure that.

sanxun0325: Yes, but in a normal registration instance scenario there will be no concurrent registration duplicate instances。You can add locks if you're worried about concurrency, but I don't think it's necessary

PR #289, opened 2021-08-05
fix leak of goroutine

fadding:
ConcurrentMap 数量并不一定恰好为SHARD_COUNT，如果不是SHARD_COUNT的情况下，会导致goroutine一直卡在wait上面不会退出，造成goroutine泄漏

### DIFF
--- a/clients/cache/concurrent_map.go
+++ b/clients/cache/concurrent_map.go
@@ -197,9 +197,9 @@ func (m ConcurrentMap) IterBuffered() <-chan Tuple {
 // It returns once the size of each buffered channel is determined,
 // before all the channels are populated using goroutines.
 func snapshot(m ConcurrentMap) (chans []chan Tuple) {
-	chans = make([]chan Tuple, SHARD_COUNT)
+	chans = make([]chan Tuple, len(m))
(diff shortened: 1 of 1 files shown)


clients/cache/concurrent_map.go
```diff
@@ -197,9 +197,9 @@ func (m ConcurrentMap) IterBuffered() <-chan Tuple {
 // It returns once the size of each buffered channel is determined,
 // before all the channels are populated using goroutines.
 func snapshot(m ConcurrentMap) (chans []chan Tuple) {
-	chans = make([]chan Tuple, SHARD_COUNT)
+	chans = make([]chan Tuple, len(m))
```
sanxun0325, line 200: In NewConcurrentMap() create len and cap equal to share count,len(m) is always equal to share count.
So it doesn't really make sense to do that, right

fadding: I encountered a goroutine leak because of this

sanxun0325: > I encountered a goroutine leak because of this

你是如何使用的，正常情况下ConcurrentMap的len和cap就是SHARD_COUNT。

Issue #259, opened 2021-06-08
client supports TLS


sanxun0325:
HTTPS mode supports tls (load certificate, parse certificate, etc.)

PR #290, opened 2021-08-06
support tls

wangweizZZ:
**purpose**
issue #259

**changelog:**
add tls.go to provide tls configuration (also applicable to grpc)

**others:**
after this part is passed, filewatcher can be added to monitor tls file change


sanxun0325 (approved):
LGTM

PR #295, opened 2021-08-19
fix: init NamingClient.INacosClient

gu18168:
In the master branch, INacosClient in NamingClient is not initialized. We should initialize it as we did in v2.

Ref: https://github.com/nacos-group/nacos-sdk-go/blob/2.0.0/clients/naming_client/naming_client.go#L44


sanxun0325 (approved):
LGTM

Issue #291, opened 2021-08-11
Supports type setting when publishing files


sanxun0325:
reference:
![image](https://user-images.githubusercontent.com/28588342/128966835-9462602d-a4d4-4bf5-b9b3-74967dfec7b9.png)

PR #296, opened 2021-08-19
support config type

wangweizZZ:
#291

### DIFF
--- a/vo/config_param.go
+++ b/vo/config_param.go
@@ -19,10 +19,10 @@ package vo
 type ConfigType string
 
 const (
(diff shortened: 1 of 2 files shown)


vo/config_param.go
```diff
@@ -16,13 +16,26 @@
 
 package vo
 
+type ConfigType string
+
+const (
+	Properties ConfigType = "properties"
```
sanxun0325, line 22: constant names should be all uppercase

wangweizZZ: ok~

sanxun0325 (approved):
LGTM

PR #327, opened 2021-11-13
client automatically re-registers

shuzhongling:
this pull request solved the issue #325 and fix a code bug when using lock

the cause of issue #325 is:
when the nacos server startup, it need a while to get ready to serve request.  if a request going to the server in this period, the server whill return 3xx.  this cause the client healthCheck failed and cause the client get a connection to the server failed.


shuzhongling:
能看出是什么地方检查失败了吗？

sanxun0325 (approved):
LGTM

PR #329, opened 2021-11-19
deps: upgrade jsonparser

shenqidebaozi:
CVE-2020-10675
moderate severity
Vulnerable versions: < 1.0.0
Patched version: 1.0.0
The Library API in buger jsonparser through 2019-12-04 allows attackers to cause a denial of service (infinite loop) via a Delete call.


sanxun0325 (approved):
LGTM

PR #338, opened 2021-11-29
fix reconnect can't subscribe problem.

horizonzy:
If send rpc request failed, it won't cache subscribe info.  when reconnect, it's won't resubscribe again.

So cache subscribe info firstly, then send rpc request.


sanxun0325 (approved):
LGTM

PR #340, opened 2021-12-03
when disconnect, sleep 1 second to avoid flush too many logs.

horizonzy:
when disconnect with server, `streamClient.Recv()` can't sync, it will return error immediately.
So here sleep 1ms.

### DIFF
--- a/common/remote/rpc/grpc_client.go
+++ b/common/remote/rpc/grpc_client.go
@@ -165,8 +165,7 @@ func (c *GrpcClient) bindBiRequestStream(streamClient nacos_grpc_service.BiReque
 						}
 					} else {
 						logger.Infof("%s received error event, isRunning:%v, isAbandon=%v, error=%+v", grpcConn.getConnectionId(), running, abandon, err)
-						after := time.After(time.Second)
-						<-after
+						<-time.After(time.Second)
 					}
 				} else {
 					c.handleServerRequest(payload, grpcConn)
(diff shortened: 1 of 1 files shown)


common/remote/rpc/grpc_client.go
```diff
@@ -158,15 +158,15 @@ func (c *GrpcClient) bindBiRequestStream(streamClient nacos_grpc_service.BiReque
...
-							logger.Errorf("%s Request stream error, switch server,error=%+v", grpcConn.getConnectionId(), err)
+							logger.Errorf("%s Request stream error, switch server, error=%+v", grpcConn.getConnectionId(), err)
 						}
 						if atomic.CompareAndSwapInt32((*int32)(&c.rpcClientStatus), int32(RUNNING), int32(UNHEALTHY)) {
 							c.switchServerAsync(ServerInfo{}, false)
 						}
 					} else {
-						logger.Infof("%s Ignore event,isRunning:%v,isAbandon=%v", grpcConn.getConnectionId(), running, abandon)
+						logger.Infof("%s received error event, isRunning:%v, isAbandon=%v, error=%+v", grpcConn.getConnectionId(), running, abandon, err)
+						time.Sleep(time.Second)
```
sanxun0325, line 168: 使用time.after 更好一点

horizonzy: 好的。

sanxun0325: <-time.After(time.Second) 大佬这样就好～

horizonzy: 好的。

binbin0325 (approved):
LGTM

PR #351, opened 2021-12-15
Fix type assertion error

JiankunDai:


sanxun0325 (commented):
Client creation through Properties will not continue to be supported in 2.0。

PR #355, opened 2021-12-20
bugfix: throw NewConfigProxy error

Gerrard-YNWA:
Currently when [`NewConfigProxy`](https://github.com/nacos-group/nacos-sdk-go/blob/2.0.0/clients/config_client/config_client.go#L100) returns an error , `NewConfigClient` just ignored it which may cause unexpected error. 

For example, an I/O timeout error in the login method, and the next request will fail due to the lack of accessToken.


sanxun0325 (approved):
LGTM

Issue #377, opened 2022-01-20
branch 2.0.0: PublishConfig  panic when config a invalid server


believening:
I start a local server which listen on 18848, no default 8848, run example config code will be panic.

```
➜  nacos-go-sdk git:(up-v2) go run example/config/main.go 
2022/01/20 17:06:10 [INFO] logDir:</tmp/nacos/log>   cacheDir:</tmp/nacos/cache>
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x9d1bc3]

goroutine 1 [running]:
github.com/nacos-group/nacos-sdk-go/v2/clients/config_client.(*ConfigClient).PublishConfig(0xc00016a820, {{0xb25d72, 0x9}, {0xb2678a, 0xa}, {0xb2779f, 0xc}, {0x0, 0x0}, {0x0, ...}, ...})
        /root/codes/slime/nacos-go-sdk/clients/config_client/config_client.go:231 +0x6a3
main.main()
        /root/codes/slime/nacos-go-sdk/example/config/main.go:60 +0x438
exit status 2
```

believening:
https://github.com/nacos-group/nacos-sdk-go/blob/e4f5d94918b521f918beb6e86fa8e79d2b91fa96/clients/config_client/config_client.go#L231

believening:
https://github.com/nacos-group/nacos-sdk-go/blob/e4f5d94918b521f918beb6e86fa8e79d2b91fa96/clients/config_client/config_client.go#L230-L231

```
func (cp *ConfigProxy) requestProxy(rpcClient *rpc.RpcClient, request rpc_request.IRequest, timeoutMills uint64) (rpc_response.IResponse, error) {
        // ......
	return rpcClient.Request(request, int64(timeoutMills))
}
```
the result of rpcClient.Request() can be nil

PR #378, opened 2022-01-20
fix a panic when requestProxy returns a nil response response when calling PublishConfig

believening:
#377


sanxun0325 (commented):
Other interfaces in the config center also have this problem, can you fix them together

believening:
> Other interfaces in the config center also have this problem, can you fix them together

I have made a new fix, please check again.

sanxun0325 (approved):
LGTM

Issue #373, opened 2022-01-17
Feature request: support output logs to stdout


wangrzneu:
Current SDK only support output log to files. Application deployed in Kubernetes or Docker environment usually output logs to stdout, and manage logs using other components. For example, https://docs.docker.com/config/containers/logging/

sanxun0325:
Can you describe in detail what the current SDK logging does not meet the requirements

wangrzneu:
@sanxun0325 I update the description and create a PR. Please help to review it.

spacewander:
@binbin0325 
It seems this feature is not completed. Why close the issue?

PR #385, opened 2022-01-25
support output log to stdout

wangrzneu:
solve https://github.com/nacos-group/nacos-sdk-go/issues/373

### DIFF
--- a/common/logger/logger.go
+++ b/common/logger/logger.go
@@ -45,6 +45,7 @@ type Config struct {
 	Level            string
 	Sampling         *SamplingConfig
 	LogRollingConfig *lumberjack.Logger
+	LogStdout        bool // whether to output log to stdout. If it is true, LogRollingConfig is ignored
 }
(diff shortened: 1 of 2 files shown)


common/logger/logger.go
```diff
@@ -174,5 +176,9 @@ func GetLogger() Logger {
 
 // getLogWriter get Lumberjack writer by LumberjackConfig
 func (c *Config) getLogWriter() zapcore.WriteSyncer {
-	return zapcore.AddSync(c.LogRollingConfig)
+	if !c.LogStdout {
```
sanxun0325, line 179: The default is to print both to the console and to a file, which may be better to reduce user configuration。
such as:
`core := zapcore.NewCore(zapcore.NewConsoleEncoder(encoder), zapcore.NewMultiWriteSyncer(writer, zapcore.AddSync(os.Stdout)), logLevel)`

PR #404, opened 2022-02-12
fix nacos reboot, app cant re register

yifeng01:
when  nacos reboot， app  cant re register,   should modify  file "connection_event_listener.go" 
`
func (c *ConnectionEventListener) CacheInstanceForRedo(serviceName, groupName string, instance model.Instance) {
	key := util.GetGroupName(serviceName, groupName)
	getInstance, ok := c.registeredInstanceCached.Get(key)
	if !ok {
		logger.Warnf("CacheInstanceForRedo get cache instance is null,key:%s", key)
	}
	if getInstance != nil && reflect.DeepEqual(getInstance.(model.Instance), instance) {
		return
	}
	c.registeredInstanceCached.Set(key, instance)
}
`

### DIFF
--- a/clients/naming_client/naming_grpc/connection_event_listener.go
+++ b/clients/naming_client/naming_grpc/connection_event_listener.go
@@ -98,9 +98,8 @@ func (c *ConnectionEventListener) CacheInstanceForRedo(serviceName, groupName st
 	getInstance, ok := c.registeredInstanceCached.Get(key)
 	if !ok {
 		logger.Warnf("CacheInstanceForRedo get cache instance is null,key:%s", key)
-		return
 	}
-	if reflect.DeepEqual(getInstance.(model.Instance), instance) {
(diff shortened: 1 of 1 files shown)


clients/naming_client/naming_grpc/connection_event_listener.go
```diff
@@ -98,9 +98,8 @@ func (c *ConnectionEventListener) CacheInstanceForRedo(serviceName, groupName st
 	getInstance, ok := c.registeredInstanceCached.Get(key)
 	if !ok {
 		logger.Warnf("CacheInstanceForRedo get cache instance is null,key:%s", key)
```
sanxun0325, line 100: instance is no longer in the cache . no need to print the log. can you delete it

sanxun0325 (approved):
可以处理一下ci go fmt 下

PR #411, opened 2022-02-18
add logger encoding options

stulzq:
support logger output json format #395


sanxun0325 (commented):
There is pr to do this, see https://github.com/nacos-group/nacos-sdk-go/pull/385

Issue #285, opened 2021-07-26
some question for namingClient asyncUpdateService


twogoods:
1. call api `intances/list` per second will be make pressure for nacos server ？

2. in `ProcessServiceJson` function 
```
	if !ok || ok && !reflect.DeepEqual(service.Hosts, oldDomain.(model.Service).Hosts) {
		if !ok {
			logger.Info("service not found in cache " + cacheKey)
		} else {
			logger.Info("service key:%s was updated to:%s", cacheKey, util.ToJsonString(service))
		}
		cache.WriteServicesToFile(*service, hr.cacheDir)
		hr.subCallback.ServiceChanged(service)
	}
```
the `service.Hosts`  instance is same , but array order is not same , so the DeepEqual is false；I have 3 nodes in nacos cluster, api call will be loadbalancered by nginx

sanxun0325:
1. call api intances/list per second will be make pressure for nacos server ？
The impact can be ignored because the instance needs to be monitored for updates and the server is the query cache efficiency is higher

2. the service.Hosts instance is same , but array order is not same , so the DeepEqual is false；I have 3 nodes in nacos cluster, api call will be loadbalancered by nginx
The order of instances is guaranteed by the nacos server, and the order of instances of each node is not different

twogoods:
> 1. call api intances/list per second will be make pressure for nacos server ？
>    _The impact can be ignored because the instance needs to be monitored for updates and the server is the query cache efficiency is higher_
> 2. the service.Hosts instance is same , but array order is not same , so the DeepEqual is false；I have 3 nodes in nacos cluster, api _call will be loadbalancered by nginx
>    The order of instances is guaranteed by the nacos server, and the order of instances of each node is not different_

@sanxun0325  if the order of instances of each node is not different, when `namingClient.Subscribe`,I  will always get changedEvent，it's incorrect

sanxun0325:
> > 1. call api intances/list per second will be make pressure for nacos server ？
> >    _The impact can be ignored because the instance needs to be monitored for updates and the server is the query cache efficiency is higher_
> > 2. the service.Hosts instance is same , but array order is not same , so the DeepEqual is false；I have 3 nodes in nacos cluster, api _call will be loadbalancered by nginx
> >    The order of instances is guaranteed by the nacos server, and the order of instances of each node is not different_
> 
[...]

PR #419, opened 2022-03-08
optimize  the way of compare service hosts.

Sivyer9303:
when using nacos 1.x cluster.the service hosts is not the same order.
so, the old code using reflect.DeepEqual will cause unnecessary operator.
```golang
if !ok || ok && !reflect.DeepEqual(service.Hosts, oldDomain.(model.Service).Hosts) {
		if !ok {
			logger.Info("service not found in cache " + cacheKey)
		} else {
			logger.Info("service key:%s was updated to:%s", cacheKey, util.ToJsonString(service))
		}
		cache.WriteServicesToFile(*service, hr.cacheDir)
		hr.subCallback.ServiceChanged(service)
               }
```
see #285 for more info.
BTW,this is the solution of 1.x. I'll create a new pr for 2.X

PR #424, opened 2022-03-24
Change the way of compare service hosts.

Sivyer9303:
change the way of compare instance.
see #419 .
this is for v2.X to fix bug #285


Sivyer9303:
@sanxun0325 plz review this.

PR #425, opened 2022-03-26
add catalog query interface

mqshen:
add catalog query interface that return total count of services.


sanxun0325 (commented):
use GetAllServicesInfo you can get all the services
https://github.com/nacos-group/nacos-sdk-go/blob/master/clients/naming_client/naming_client.go#L194

mqshen:
this api not return total count of all services.

sanxun0325:
> 



> this api not return total count of all services.

count is all services num

mqshen:
> 

count is size of return Doms. not the all services.

Issue #432, opened 2022-04-01
change the way of create cache key when load cache from file


Sivyer9303:
When ProcessService,the way of creating cache key is different from the way of creating file name.
Below is the way of creating cache key,which is the format like 'GROUP_NAME@@SERVICE_NAME@@CLUSTER'.
`cacheKey := util.GetServiceCacheKey(util.GetGroupName(service.Name, service.GroupName), service.Clusters)`.
Below is the way of creating file name,which is the format like 'SERVICE_NAME@@CLUSTER'.
`domFileName := GetFileName(util.GetServiceCacheKey(service.Name, service.Clusters), cacheDir)`.

It'll cause some unexpected condition when load cache from file.The file cache will output to service map using the file name which is different the cache key when get service. 
the service map will preserve duplicate service info.

PR #434, opened 2022-04-03
bugfix.change the way of creating cache key when load service from di…

Sivyer9303:
bugfix.change the way of creating cache key when load service from disk.
see #432


sanxun0325 (approved):
LGTM

PR #463, opened 2022-05-16
a swich for append to stdout

robberphex:


realJackSun (approved):

SpecialYang (approved):
LGTM

common/constant/config.go
```diff
@@ -47,6 +47,7 @@ type ClientConfig struct {
 	LogDir               string                   // the directory for log, default is current path
 	LogLevel             string                   // the level of log, it's must be debug,info,warn,error, default value is info
 	ContextPath          string                   // the nacos server contextpath
+	AppendToStdout       bool                     // if append log to stdout
```
SpecialYang, line 50: The legacy behavior is default to append to stdout. Shall we also does that?

SpecialYang (changes requested):

Issue #476, opened 2022-06-15
GetService method should bind namespace param


lishuo5263:
### Use case(s) - what problem will this feature solve?
when I register tuo service,name is same  ex : serverA, but different namespace,the GetService get cache should return bouth different serverA
### Proposed Solution
refactor this method ,bind namespace to cache

```
func GetServiceCacheKey(serviceName string, clusters string) string {
	if clusters == "" {
		return serviceName
	}
	return serviceName + constant.SERVICE_INFO_SPLITER + clusters
}
```
### Alternatives Considered
Acroding to this : 

```
func GetConfigCacheKey(dataId string, group string, tenant string) string {
	return dataId + constant.CONFIG_INFO_SPLITER + group + constant.CONFIG_INFO_SPLITER + tenant
}
```
should bind tenant to use namespace

lishuo5263:
com/alibaba/nacos/naming/core/Instance.java

sanxun0325:
Different namespace should create different naming client
```go
	cc := *constant.NewClientConfig(
		constant.WithNamespaceId(), // set namespaceid
		constant.WithTimeoutMs(5000),
		constant.WithNotLoadCacheAtStart(true),
		constant.WithLogDir("/tmp/nacos/log"),
		constant.WithCacheDir("/tmp/nacos/cache"),
		constant.WithLogLevel("debug"),
	)

	// create naming client
	client, err := clients.NewNamingClient(
		vo.NacosClientParam{
			ClientConfig:  &cc,
			ServerConfigs: sc,
		},
	)
`

PR #477, opened 2022-06-15
[#476 feature: add support tenant to client query]

lishuo5263:
### Feature
* Add cache param Tenant
* Support client use namespace to query Instance

### DIFF
--- a/clients/cache/disk_cache.go
+++ b/clients/cache/disk_cache.go
@@ -45,7 +45,7 @@ func GetFileName(cacheKey string, cacheDir string) string {
 func WriteServicesToFile(service model.Service, cacheDir string) {
 	file.MkdirIfNecessary(cacheDir)
 	sb, _ := json.Marshal(service)
-	domFileName := GetFileName(util.GetServiceCacheKey(service.Name, service.Clusters), cacheDir)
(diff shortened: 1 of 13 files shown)


sanxun0325 (commented):
Why add tenant? query by NamespaceID is already available

lishuo5263:
> Why add tenant? query by NamespaceID is already available
I can't upload the screen short, i will send mail talk with you,and explain reason.

sanxun0325:
image
each namespce id corresponds to a client. The service map in the client cache service instances under the current namespace

sanxun0325:
If you only want to query the instance list based on the namespace ID, refer: https://github.com/nacos-group/nacos-sdk-go/blob/master/example/service/main.go#L271

lishuo5263:
> If you only want to query the instance list based on the namespace ID, refer: https://github.com/nacos-group/nacos-sdk-go/blob/master/example/service/main.go#L271

Em.. I thought the cc should be support entire namespace to query,not like ony bind one namespace to use.

lishuo5263:
> If you only want to query the instance list based on the namespace ID, refer: https://github.com/nacos-group/nacos-sdk-go/blob/master/example/service/main.go#L271

Em.. I thought the cc should be support entire namespace to query,not like ony bind one namespace to use.

sanxun0325:
> 

don't need to do this; GetServiceList supports passing namespaceID

PR #491, opened 2022-07-18
feature.add limiter for config client.

Sivyer9303:
Add limiter for config request to avoid dealing too much GetConfig request in nacos-server.
The trial limiter configuration is allow 5 request per second,and max reserver stock is 5 too.which is the same as Java nacos client.


sanxun0325 (approved):
LGTM

PR #497, opened 2022-08-01
 time ticker instand  sleep

jasonchuan:
time sleep is a  waste of cpu time ,so take time ticker  instand

### DIFF
--- a/clients/naming_client/service_info_updater.go
+++ b/clients/naming_client/service_info_updater.go
@@ -45,6 +45,7 @@ func NewServiceInfoUpdater(serviceInfoHolder *naming_cache.ServiceInfoHolder, up
 func (s *ServiceInfoUpdater) asyncUpdateService() {
 	sema := util.NewSemaphore(s.updateThreadNum)
 	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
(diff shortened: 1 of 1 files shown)


chuntaojun (approved):

clients/naming_client/service_info_updater.go
```diff
@@ -44,23 +44,26 @@ func NewServiceInfoUpdater(serviceInfoHolder *naming_cache.ServiceInfoHolder, up
 
 func (s *ServiceInfoUpdater) asyncUpdateService() {
 	sema := util.NewSemaphore(s.updateThreadNum)
+	ticker := time.NewTicker(time.Second)
 	for {
```
chuntaojun, line 48: add defer ticker.Stop()

binbin0325 (commented):
sleep & ticker bottom layer is timer, no difference.
how is ticker better than sleep?

jasonchuan:
> sleep & ticker bottom layer is timer, no difference. how is ticker better than sleep?
you are right , sleep func uses timer cache in g  in go 1.9 version ,  sleep would new timer instance every time called before 1.9 version . Nacos sdk need after 1.15,  so is OK.Thank You ！   
the only one difference is that  using ticker will put timer instance on stack after escape analysis  ,sleep on heap ,  ticker is little  better   than sleep for GC

binbin0325:
> 

So I don't see any improvement in pr at the moment, can I turn it off?

binbin0325:
> > sleep & ticker bottom layer is timer, no difference. how is ticker better than sleep?
> > you are right , sleep func uses timer cache in g  in go 1.9 version ,  sleep would new timer instance every time called before 1.9 version . Nacos sdk need after 1.15,  so is OK.Thank You ！
> > the only one difference is that  using ticker will put timer instance on stack after escape analysis  ,sleep on heap ,  ticker is little  better   than sleep for GC

so I don't see any improvement in pr at the moment, can I turn it off?

jasonchuan:
> 
OK

binbin0325:
> > 
> 
> OK

Issue #517, opened 2022-09-09
客户端关闭无效


binbin0325:
### What version of nacos-sdk-go are you using?
2.x
### What version of nacos-sever are you using?
2.x
### What version of Go are you using (`go version`)?
1.15
### What operating system (Linux, Windows, …) and version?
mac
### What did you do?
If possible, provide a recipe for reproducing the error.
客户端关闭后，健康检查会把状态改为UNHEALTHY，导致链接重新建立
### What did you expect to see?
客户端关闭后，链接断开

PR #522, opened 2022-09-27
Fix leaking of goroutines of naming client.

wangjian-pg:
The `naming client` creates a few background goroutines to handle service info updating, service instance heart beating, gRPC connection reconnecting, etc. Most of the created goroutines are still working even if the `naming client` has been closed. This would be a big problem if the user creates many different naming clients and the lifecycle of the clients are shorter than the program.

This PR fixed the problem of `naming client`, however,  the `config client` still has the same issue. I may propose another PR to fix the same problem of `config client` later.

BTW, #517 is somehow caused by this goroutine leaking problem. And it has been fixed by this PR too.

### DIFF
--- a/clients/naming_client/naming_client.go
+++ b/clients/naming_client/naming_client.go
@@ -39,6 +39,7 @@ type NamingClient struct {
 	nacos_client.INacosClient
 	serviceProxy      naming_proxy.INamingProxy
 	serviceInfoHolder *naming_cache.ServiceInfoHolder
+	svcInfoUpdater    *ServiceInfoUpdater
 }
 
 // NewNamingClient ...
@@ -70,11 +71,18 @@ func NewNamingClient(nc nacos_client.INacosClient) (*NamingClient, error) {
 		clientConfig.UpdateCacheWhenEmpty, clientConfig.NotLoadCacheAtStart)
 
 	naming.serviceProxy, err = NewNamingProxyDelegate(clientConfig, serverConfig, httpAgent, naming.serviceInfoHolder)
-
-	go NewServiceInfoUpdater(naming.serviceInfoHolder, clientConfig.UpdateThreadNum, naming.serviceProxy).asyncUpdateService()
 	if err != nil {
 		return naming, err
 	}
+
+	naming.svcInfoUpdater = NewServiceInfoUpdater(naming.serviceInfoHolder,
+		clientConfig.UpdateThreadNum, naming.serviceProxy)
+
+	go func() {
+		defer logger.Infof("goroutine for updating service of client: <%s:%s:%s> exit",
+			clientConfig.NamespaceId, clientConfig.AppName, clientConfig.AppKey)
+		naming.svcInfoUpdater.asyncUpdateService()
+	}()
 
 	return naming, nil
 }
@@ -309,4 +317,5 @@ func (sc *NamingClient) Unsubscribe(param *vo.SubscribeParam) (err error) {
 // CloseClient ...
 func (sc *NamingClient) CloseClient() {
 	sc.serviceProxy.CloseClient()
+	sc.svcInfoUpdater.Close()
 }
--- a/clients/naming_client/naming_http/beat_reactor.go
+++ b/clients/naming_client/naming_http/beat_reactor.go
@@ -47,11 +47,11 @@ type BeatReactor struct {
 	beatRecordMap       cache.ConcurrentMap
 	clientCfg           constant.ClientConfig
 	mux                 *sync.Mutex
+	cancelBeating       func()
+	beatingCtx          context.Context
 }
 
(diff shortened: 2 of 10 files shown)


binbin0325:
Users should not create clients more than once. client should be global. 
issue https://github.com/nacos-group/nacos-sdk-go/issues/517 ,It is the health check that causes the link to be re-created

wangjian-pg:
We used the SDK at a Cloud infrastructure to sync the Nacos registered services to Istio. Users can choose the Nacos that they want to sync and stop the syncing at any time. After the syncing stopped, we closed the `NamingClient` and found that a few of the goroutines leaked.

It's a common case that a client is created, does something, and finally, the client is closed, as we usually do with network connections and filesystems. Maybe we just want to take a look at the registered services and do nothing other with the client anymore, leaving all the goroutines running in the background is wasteful in this scenario. After all, resources leaking after the object is closed is very unexpected, IMO.

As for issue #517, it is the leaked goroutine used for reconnecting the gRPC stream and health checking after the client is closed that makes the reconnecting happen, and if the corresponding goroutine exit,  this should never happen.

https://github.com/nacos-group/nacos-sdk-go/blob/c65ee73bd7c8bd569a503da2db234db4bcd58849/common/remote/rpc/rpc_client.go#L178-L205

binbin0325:
Yes, there will be some optimizations later. before do that, need to tease out which goroutine leaks after the shutdown。

wangjian-pg:
I've checked all goroutines created by `NamingClient` and made sure that all created goroutines will exit after the client is closed. Please review this PR. Up for suggestions.

binbin0325:
Maybe you can use one context to cancel all goroutines

wangjian-pg:
Yes, I may. But passing one context to all of the goroutines created by different modules would cause a lot code changes. Since most of the relevant structs already have a `Close` method, using a `stopChan` to instruct the goroutine to exit seems like a more straightforward way.

binbin0325:
but the current implementation, it's not elegant, and it's heavily modified

wangjian-pg:
I have considered different approaches and finally chose this one as it changes less code than the others. I have already run a test locally and made sure that all of the goroutines created by `Naming Client` exit after the client closed. 

I will keep optimizing the code, bit by bit,  and a more elegant way should come up. As for now, I think the most important thing is to fix the problem ASAP.

binbin0325:
Elegant,correct code is more important in open source projects



---- Replied Message ----
| From | ***@***.***> |
| Date | 09/28/2022 21:36 |
| To | ***@***.***> |
| Cc | ***@***.******@***.***> |
| Subject | Re: [nacos-group/nacos-sdk-go] Fix leaking of goroutines of naming client. (PR #522) |

I have considered different approaches and finally chose this one as it changes less code than the others. I have already run a test locally and made sure that all of the goroutines created by Naming Client exit after the client closed.

I will keep optimizing the code, bit by bit, and a more elegant way should come up. As for now, I think the most important thing is to fix the problem ASAP.

wangjian-pg:
I do not think using one context to cancel all of the created goroutines is a good idea.
1. Just as the doc of `context.Context` said,  `Context` is a request-scoped object and we usually use it to control the lifecycle of a request/API, just like `http.Request.WithContext` do. This is not the case we try to resolve here.
image
2.  One of the critical things that `Context`'s cancel function actually do is close the `Context`'s `done` channel, just the same idea as I used in this PR. 

Introducing the `context.Context` to control the lifecycle of all created background goroutines not seems like a good idea. It would cause a lot of code refactoring and not that idiomatic.

binbin0325:
The context is not only used in the http request

wangjian-pg:
The `http.Request` is just an example to show that `conext.Context` is usually request-scoped😊.  Just like a doc of package(https://pkg.go.dev/context) `context` said: 

> Incoming requests to a server should create a Context, and outgoing calls to servers should accept a Context. The chain of function calls between them must propagate the Context, optionally replacing it with a derived Context created using WithCancel, WithDeadline, WithTimeout, or WithValue. When a Context is canceled, all Contexts derived from it are also canceled.


> Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.

binbin0325:
It's clear what the context does, there's nothing to talk about here, the context doesn't just apply to the request, it's just an example of using it in a Web request, our goal is to cancel all child goroutines, it also sends data to chan, Using it to implement Goroutine that cancels all leaks is relatively simple and clear.

binbin0325:
I will submit the PR later to improve it. If you are interested, you can also modify the current PR

wangjian-pg:
Looking forward to your optimization.

wangjian-pg:
Close the PR for not agreeing. I will keep optimizing the code in my forked repo and give feedback if I find any bugs.

PR #559, opened 2023-01-16
a timer make code more robust

haifeiWu:
use a timer replace time.Sleep(1 * time.Second) make code rebust


clients/naming_client/service_info_updater.go
```diff
@@ -47,11 +47,13 @@ func NewServiceInfoUpdater(ctx context.Context, serviceInfoHolder *naming_cache.
 
 func (s *ServiceInfoUpdater) asyncUpdateService() {
 	sema := util.NewSemaphore(s.updateThreadNum)
+	duration := 1 * time.Second
+	timer := time.NewTimer(duration) // new timer
```
binbin0325, line 51: The underlying implementation of timer is no different from sleep

binbin0325 (changes requested):

PR #575, opened 2023-02-16
Fix MaxInt64 overflows on ARM 32-bits

bavelee:
`math.MaxInt64` is untyped constant. This causes overflow on a 32-bit system.

Reproduce:
```bash
$ GOARCH=arm go build example/config/main.go 
go: downloading golang.org/x/net v0.0.0-20210525063256-abc453219eb5
go: downloading golang.org/x/sys v0.0.0-20220114195835-da31bd327af9
go: downloading golang.org/x/text v0.3.6
# github.com/nacos-group/nacos-sdk-go/v2/clients/config_client
clients\config_client\config_client.go:141:45: math.MaxInt64 (untyped int constant 9223372036854775807) overflows int
```

Edit:
`clients\config_client\config_client.go:141:45`

```go
max := int64(math.MaxInt64)
```

### DIFF
--- a/clients/config_client/config_client.go
+++ b/clients/config_client/config_client.go
@@ -138,9 +138,8 @@ func NewConfigClient(nc nacos_client.INacosClient) (*ConfigClient, error) {
 
 	config.cacheMap = cache.NewConcurrentMap()
 	// maximum buffered queue to prevent chan deadlocks during frequent configuration file updates
-	// fix overflows on ARM 32-bits
-	max := int64(math.MaxInt64)
-	config.listenExecute = make(chan struct{}, max)
+	// use Math.MaxInt to avoid overflows on ARM 32-bits
+	config.listenExecute = make(chan struct{}, math.MaxInt)
 
 	config.startInternal()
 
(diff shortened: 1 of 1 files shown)


clients/config_client/config_client.go
```diff
@@ -138,7 +138,9 @@ func NewConfigClient(nc nacos_client.INacosClient) (*ConfigClient, error) {
 
 	config.cacheMap = cache.NewConcurrentMap()
 	// maximum buffered queue to prevent chan deadlocks during frequent configuration file updates
-	config.listenExecute = make(chan struct{}, math.MaxInt64)
```
binbin0325, line 141: math.MaxInt may be better

bavelee:
i have tested with `math.MaxInt` and it works fine

clients/config_client/config_client.go
```diff
@@ -138,7 +138,8 @@ func NewConfigClient(nc nacos_client.INacosClient) (*ConfigClient, error) {
...
+	// use Math.MaxInt to avoid overflows on ARM 32-bits
+	config.listenExecute = make(chan struct{}, math.MaxInt)
```
binbin0325, line 142: Sorry, maxInt is only supported in golang 1.17, so use maxInt32 instead

binbin0325 (changes requested):

clients/config_client/config_client.go
bavelee: OK, I changed it to MaxInt and made corresponding tests and it works normally

binbin0325 (approved):

PR #605, opened 2023-04-20
Update object2param.go

miajio:
As a public method, it needs to be compatible when there is another object in the passed structure, then the previous method cannot be compatible

### DIFF
--- a/util/object2param.go
+++ b/util/object2param.go
@@ -58,7 +58,7 @@ func TransformObject2Param(object interface{}) (params map[string]string) {
 					if !valueOf.Field(i).IsNil() {
 						bytes, err := json.Marshal(valueOf.Field(i).Interface())
 						if err != nil {
-							// logger.Errorf("[TransformObject2Param] json.Marshal err:%+v", err)
+							logger.Errorf("[TransformObject2Param] json.Marshal err:%+v", err)
 						} else {
 							params[tag] = string(bytes)
 						}
--- a/util/object2param.go
+++ b/util/object2param.go
@@ -25,6 +25,7 @@ import (
 	"github.com/nacos-group/nacos-sdk-go/v2/common/logger"
 )
 
+// TransformObject2Param method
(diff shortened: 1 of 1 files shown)


binbin0325 (changes requested):
Why make this change? Please provide a detailed description

miajio:
I have reviewed your testing method, but please take a look at my testing method below. If the user uses a * time. Time when using this method, the original parsing method will not pass, and subsequent default conditions will be added. Therefore, the user only needs to implement String() on their custom structure to use this common method.

oh i'm sorry, At the same time, based on the submission, I annotated an existing log print, which will result in an invalid import

type object struct {
	Name     string            `param:"name"`
	Likes    []string          `param:"likes"`
	Metadata map[string]string `param:"metadata"`
	Age      uint64            `param:"age"`
	Healthy  bool              `param:"healthy"`
	Money    int               `param:"money"`
	Now      *time.Time        `param:"now"`
}

func TestTransformObject2Param(t *testing.T) {
	assert.Equal(t, map[string]string{}, TransformObject2Param(nil))

	nowTime := time.Now()

	obj := object{
		Name:  "code",
		Likes: []string{"a", "b"},
		Metadata: map[string]string{
			"M1": "m1",
		},
		Age:     10,
		Healthy: true,
		Money:   10,
		Now:     &nowTime,
	}
	params := TransformObject2Param(&obj)
	assert.Equal(t, map[string]string{
		"name":     "code",
		"metadata": `{"M1":"m1"}`,
		"likes":    "a,b",
		"age":      "10",
		"money":    "10",
		"healthy":  "true",
		"now":      nowTime.String(),
	}, params)
}

binbin0325:
> I have reviewed your testing method, but please take a look at my testing method below. If the user uses a * time. Time when using this method, the original parsing method will not pass, and subsequent default conditions will be added. Therefore, the user only needs to implement String() on their custom structure to use this common method.

`util.TransformObject2Param()` just for nacos-go-sdk for internal use, if you need to use this ability can be realized.

PR #622, opened 2023-06-18
return error if ErrorCode is 300

xpume:
By failing fast, developers learn what won't work and can quickly move on to a better approach.


binbin0325 (changes requested):
https://github.com/nacos-group/nacos-sdk-go/pull/626 和这个pr是同一件事情

PR #652, opened 2023-08-30
log it when response is fail and no error

sodaRyCN:
[653](https://github.com/nacos-group/nacos-sdk-go/issues/653)

### DIFF
--- a/common/remote/rpc/rpc_client.go
+++ b/common/remote/rpc/rpc_client.go
@@ -506,7 +506,7 @@ func (r *RpcClient) Request(request rpc_request.IRequest, timeoutMills int64) (r
 			continue
 		}
 		if response != nil && !response.IsSuccess() {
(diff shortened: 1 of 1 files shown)


common/remote/rpc/rpc_client.go
```diff
@@ -505,6 +505,9 @@ func (r *RpcClient) Request(request rpc_request.IRequest, timeoutMills int64) (r
 			currentErr = waitReconnect(timeoutMills, &retryTimes, request, errors.New(response.GetMessage()))
 			continue
 		}
+		if response != nil && !response.IsSuccess() {
+			logger.Warnf("%s request received fail response, error code %d, result code %d, message %s", request.GetRequestType(), response.GetErrorCode(), response.GetResultCode(), response.GetMessage())
```
binbin0325, line 509: modify the print format, the parameter with =or: link, such as: `errCode=%d, resultCode=%d`

PR #656, opened 2023-09-06
support KMSv3 encrypt/decrypt mod

robynron:
support KMSv3 encrypt/decrypt mod


binbin0325 (approved):

robynron:
still some fixes need to be done

robynron:
Using `transferKmsClient` instand of `kmsClient` to fit both kmsv1 and kmsv3 and both of the kms mods examples supplied.

PR #686, opened 2023-11-28
fix:path separator conversion for different platforms

iTuringTR:
when I input a dir as "./nacos/cache" on windows,I encountered an error:"mkdir d:\goProject\api_gw\test\./nacos/cache: The system cannot find the path specified."

### DIFF
--- a/common/file/file.go
+++ b/common/file/file.go
@@ -39,7 +39,14 @@ func init() {
 }
 
 func MkdirIfNecessary(createDir string) (err error) {
-	s := strings.Split(createDir, path)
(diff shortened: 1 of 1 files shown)


binbin0325 (commented):
In the windows environment, you should use \ when filling in the directory. so there's no need to change it here

iTuringTR:
> In the windows environment, you should use \ when filling in the directory. so there's no need to change it here

oK,I think is that maybe when the user transplants the code from the windwos to the linux platform, an error may be reported because he forgets to modify the path....hahaha...

PR #699, opened 2023-12-15
[version compatibility] fix up when response.ResultCode is 200 but response.Success is false

### LINKED ISSUE #698: [version compatibility] success field of nacos server's response is empty
v2.2.2

2.0.3

go version go1.21.5 darwin/arm64

### What did you do?
image

As you can see, there is no `field` for `success` in `response`, resulting in `unmarshal` to `success` in response being set to `false`. However, `resultCode` is 200, which means that the result of the request is successful.

-------

可以看到，`response`中没有`success`这个`field`，导致`unmarshal`到response中的`success`被设置成`false`。但是`resultCode`为`200`，表示这个请求的结果是成功的。

robynron:
fix #698


robynron:
image

KomachiSion (approved):

PR #739, opened 2024-03-29
fix panic when server push request has request headers.

brotherlu-xcq:
check if the headers are nil before iput all headers to request.


binbin0325 (approved):
LGTM

PR #740, opened 2024-03-31
🐞 fix:Adjust lock order to address concurrency issue

XiaoK29:
Adjust lock order to address concurrency issue  Moved defer ed.mux.Unlock() after ed.mux.Lock() to ensure immediate unlocking after acquiring the lock, resolving potential concurrency issues.


binbin0325 (approved):
LGTM

PR #779, opened 2024-10-11
fix: md5 not empty when content is empty

EndlessSeeker:
- Md5 is not empty when the content of dataId is empty, which causes high frequency calls for nacos server after dataId is deleted.


binbin0325 (approved):
LGTM

PR #783, opened 2024-11-07
Update nacos_grpc_service.proto

yanxyc:
need go_package to void  conflict, maybe need regenerate .pb.go file

panic: proto: file "nacos_grpc_service.proto" has a name conflict over Request
See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict


binbin0325 (approved):
LGTM

PR #799, opened 2025-02-24
Support publish config with config_tags

Sunrisea:
Support publish config with config_tags, add ConfigTags in ConfigParam


binbin0325 (approved):
LGTM

PR #802, opened 2025-03-11
fix(#790): resubscribe to previously subscribed configurations after rpcClient reconnects to the serve

Bo-Qiu:
fix(#790 ): resubscribe to previously subscribed configurations after rpcClient reconnects to the server


Bo-Qiu:
Hi, please help review the code, thx!

@binbin0325

PR #827, opened 2025-05-27
fix: update the mock package

saltbo:
It seems that the mock package has not been updated for a long time and is no longer consistent with the interface.


binbin0325 (approved):

saltbo:
Can this be merged?

PR #846, opened 2025-09-12
Fix the IP and port verification logic during service registration

lengyue1084:
-Added legitimacy verification for IP addresses and domain names
-Added validity check for port numbers to ensure they are within the valid range


Sunrisea:
The IP field allows users to enter a domain name. Is it appropriate to perform format validation here?

PR #876, opened 2026-04-23
fix: handle clients/config_client errors

alrs:
This picks up four unhandled `err` variables in `clients/config_client`.

### DIFF
--- a/clients/config_client/config_proxy.go
+++ b/clients/config_client/config_proxy.go
@@ -134,8 +134,14 @@ func (cp *ConfigProxy) queryConfig(dataId, group, tenant string, timeout uint64,
 		return nil, errors.New("ConfigQueryRequest returns type error")
 	}
 	if response.IsSuccess() {
-		cache.WriteConfigToFile(cacheKey, cp.clientConfig.CacheDir, response.Content)
-		cache.WriteEncryptedDataKeyToFile(cacheKey, cp.clientConfig.CacheDir, response.EncryptedDataKey)
+		err = cache.WriteConfigToFile(cacheKey, cp.clientConfig.CacheDir, response.Content)
+		if err != nil {
+			return nil, err
+		}
+		err = cache.WriteEncryptedDataKeyToFile(cacheKey, cp.clientConfig.CacheDir, response.EncryptedDataKey)
+		if err != nil {
+			return nil, err
+		}
 		if response.ContentType == "" {
 			response.ContentType = "text"
 		}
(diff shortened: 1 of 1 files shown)


Sunrisea:
Thanks @alrs for catching these unhandled errors — these are real linter findings worth fixing.

**What looks good**
- All 4 unhandled errors accurately located
- The `errorCode=300` (config-not-found) branch is also covered, nice consistency

**Suggested direction** (please adjust before merge)

In this SDK, `cache.WriteConfigToFile` / `cache.WriteEncryptedDataKeyToFile` play a **fallback-snapshot** role: they exist so that when the server is unreachable and `NotLoadCacheAtStart=false`, the client can recover from the on-disk snapshot via `config_client.go::loadCacheCallback`. They are not part of the success semantics of `GetConfig` itself.

The current change promotes a snapshot-write failure to a `GetConfig` failure, which has two side effects:

1. **Hidden breaking change.** When the disk is full or unwritable, the server has already returned the correct config, yet the caller receives an error. This contradicts the existing SDK contract that cache is a network-fallback mechanism and must not break the main read path.
2. **Semantics collapse on `errorCode=300`.** Originally this branch communicates "config does not exist". After this change, a disk error masks that signal and the caller can no longer distinguish "not found" from "snapshot write failed".

Also note that `disk_cache.go::WriteConfigToFile` already logs the failure internally via `logger.Error`, so propagating it out causes the same failure to be logged twice.

A **soft-fail** approach satisfies the linter while preserving the snapshot semantics:

```go
if response.IsSuccess() {
    if err := cache.WriteConfigToFile(cacheKey, cp.clientConfig.CacheDir, response.Content); err != nil {
        logger.Errorf("[config_rpc_client] write config snapshot failed, dataId=%s, group=%s, tenant=%s, err=%v",
            dataId, group, tenant, err)
    }
    if err := cache.WriteEncryptedDataKeyToFile(cacheKey, cp.clientConfig.CacheDir, response.EncryptedDataKey); err != nil {
        logger.Errorf("[config_rpc_client] write encrypted data key snapshot failed, dataId=%s, group=%s, tenant=%s, err=%v",
            dataId, group, tenant, err)
    }
    if response.ContentType == "" {
        response.ContentType = "text"
    }
    return response, nil
}

if response.GetErrorCode() == 300 {
    if err := cache.WriteConfigToFile(cacheKey, cp.clientConfig.CacheDir, ""); err != nil {
        logger.Errorf("[config_rpc_client] clear config snapshot failed, dataId=%s, group=%s, tenant=%s, err=%v",
            dataId, group, tenant, err)
    }
    if err := cache.WriteEncryptedDataKeyToFile(cacheKey, cp.clientConfig.CacheDir, ""); err != nil {
        logger.Errorf("[config_rpc_client] clear encrypted data key snapshot failed, dataId=%s, group=%s, tenant=%s, err=%v",
            dataId, group, tenant, err)
    }
    response.SetSuccess(true)
    return response, nil
}
```

Could you adjust the PR along these lines? Once it's soft-fail we'll be happy to merge. Happy to discuss if you have concerns. Thanks!

Issue #898, opened 2026-07-20
PR3: Auth layer — expose login errors, TTL-based token refresh, concurrency safety


cxhello:
Part of the v3 upgrade roadmap (#879), PR3 of 7. Depends on PR1 (#881) and PR2 (#897), both merged.

## Scope — Auth layer

Refactor `common/security` to expose login failures, refresh tokens by TTL, and eliminate data races, while preserving the existing `SecurityProxy`/`AuthClient` type names (incremental migration per the roadmap guidance).

- **Expose login errors (#670):** `SecurityProxy.Login()` returns an aggregated `error`; new `ClientConfig.FailOnAuthError` (default `false`) makes `NewClient` fail fast on credential errors (HTTP 401/403) while transient/network errors keep retrying. Level-graded logging so even without the flag, users can tell "wrong password" from "server unreachable".
- **TTL-based refresh:** `AuthClient` gains `AutoRefresh(ctx)`; each client self-schedules (`NacosAuthClient` by token TTL, `RamAuthClient` no-op). Removes `SecurityProxy`'s unconditional 5s polling loop.
- **Concurrency safety:** guard `NacosAuthClient`'s refresh state (`lastRefreshTime`/`tokenTtl`/`tokenRefreshWindow`/`serverCfgs`) with a mutex (`accessToken` stays `atomic.Value`); no lock held across HTTP IO.
- **#807 / #728:** core already fixed on the branch by prior commits (token refresh: f1545e0; endpoint-mode login: #730); this PR adds regression tests and error exposure, then closes them.

## Related
- Fixes #670
- Refs #807, #728 (regression coverage; closed on merge)
- Refs #879 (roadmap)

cxhello:
Done in #899, merged into `v3.x-dev` as 47d46c78c59cec6147296fb5ef3f344fcc2f9551.

Since the PR targeted `v3.x-dev` rather than the default branch, the `Fixes` keyword did not auto-close this sub-issue, so closing it manually (same as #880 and #896).

The upstream issues #807 / #670 / #728 are left open for the maintainers to close at their discretion.

PR #897, opened 2026-07-17
feat: wire PayloadCodec and nacos-sdk-proto into gRPC request/response/server-push paths

cxhello:
Refs #879, closes #896. Partially addresses #654 (see Known limitation).

## Summary

Second PR of the v3 roadmap (#879): wire `PayloadCodec` and `nacos-sdk-proto` message types into the actual gRPC request, response, and server-push paths, per the scope agreed in the [PR1 progress update](https://github.com/nacos-group/nacos-sdk-go/issues/879#issuecomment-4994983661).

### Changes

1. **Transport type replacement** (`1c3042a`) — the locally generated stubs (`api/grpc`) are replaced with the `nacos-sdk-proto/go` equivalents, and `api/proto/` + `api/grpc/` are deleted **in the same commit**. This atomicity is a hard constraint, not cleanup: both descriptors register root-namespace `Payload`/`Metadata` (the proto file has no `package`), so linking both into one binary panics at init with a duplicate protobuf registration.
2. **Codec serialization options** (`a1e47f7`) — `Encode` uses `EmitDefaultValues: true` (per the cross-language zero-value finding in alibaba/nacos#14683: omitting zeros makes the server unable to tell unset from zero); `Decode` uses `DiscardUnknown: true` (servers serialize Java getter-derived fields such as `success` that do not exist in the proto definitions, so strict decoding fails).
3. **Send path** (`c1d7a08`, `de1ead9`) — new `codec.ProtoConvertible` interface; `HealthCheckRequest`/`ServerCheckRequest`/`ConnectionSetupRequest` carry sdk-proto messages and are encoded through `PayloadCodec`; everything else stays on the legacy JSON path (with fallback on encode error). Both paths produce the identical wire format `Payload{Metadata{type}, body: Any{value: JSON bytes}}`.
4. **Response decode** (`1f6338f`) — `HealthCheckResponse`/`ServerCheckResponse`/`ErrorResponse` decode through `PayloadCodec` and are adapted back into the legacy `rpc_response` structs, so downstream type assertions keep working. `success` derivation follows the existing `InnerResponseJsonUnmarshal` rule (explicit wire field wins, else `resultCode==200`). Migrated types fail fast on decode error. Note: because `ErrorResponse` is migrated, error responses of **unmigrated** request types (e.g. a failed `ConfigQueryRequest`) also take the proto decode path — the adapted struct state is field-identical to the legacy `json.Unmarshal` result, verified by unit tests.
[...]

Issue #879, opened 2026-05-22
[Proposal] nacos-sdk-go v3: Adopt nacos-sdk-proto + gRPC Layer Refactor


cxhello:
## Background

nacos-sdk-go (v2.3.5) uses gRPC as its sole transport protocol but has several long-standing issues:

### Current Problems

| Category | Issues | Impact |
|----------|--------|--------|
| gRPC Layer Bugs | #863 reconnect deadlock, #861 concurrent map writes, #833 data race, #654 proto registration panic | Production crashes and hangs |
| Hand-written Message Types | Request/response structs manually maintained in `rpc_request/` and `rpc_response/` | Drift from Java server definitions, missing fields |
| Auth Issues | #807 token never refreshes, #670 login failures silently ignored | Connection drops in long-running services |
| Naming Bugs | #866 registration overwrites, #865 Nacos 3.x subscribe broken, #780 redo key too coarse | Service discovery reliability |
| Config Bugs | #694 notifications lost after restart, #629 unsubscribe broken, #727 CAS silent failure | Configuration consistency |
| Missing Features | No distributed lock (#864), no FuzzyWatch (#859), no AI/MCP support | Nacos 3.x feature gap |
| Outdated Dependencies | #871 alibaba-cloud-sdk-go v1 (unmaintained), jaeger-client-go (deprecated) | Security and maintenance risk |
| Weak CI | Only `gofmt` + `go test`, no integration tests, no Nacos Server containers | Low test coverage |

### nacos-sdk-proto

[nacos-group/nacos-sdk-proto](https://github.com/nacos-group/nacos-sdk-proto) (v1.0.0-beta.9, see alibaba/nacos#14683) now provides unified proto definitions with generated Go code:

- 42 proto files, 112 messages across 5 modules: common, config, naming, lock, ai
- Go package: `github.com/nacos-group/nacos-sdk-proto/go`
- Already adopted by [nacos-sdk-nodejs PR #146](https://github.com/nacos-group/nacos-sdk-nodejs/pull/146)

## Proposal

Upgrade nacos-sdk-go from v2 to **v3**, adopting `nacos-sdk-proto/go` as the message type source and refactoring the gRPC layer for reliability.

### Key Changes

1. **Replace hand-written structs** with proto-generated types via `nacos-sdk-proto/go`
2. **Implement PayloadCodec** — centralized message type registry + protojson encode/decode
[...]

Issue #833, opened 2025-06-26
🐞 Bug Report: Data Race in RpcClient.Start() and sendHealthCheck()


lengpucheng:
**Describe the bug**

A data race was detected between RpcClient.Start() and the internal sendHealthCheck() goroutine in version v2.3.1 of nacos-sdk-go.

The issue occurs when a field inside RpcClient is accessed concurrently without proper synchronization: one goroutine is writing the field inside Start(), while another is reading it in sendHealthCheck().

**To Reproduce**

Run a simple client that initializes a naming client using NewNamingClient() while using the -race flag:
```go
if utils.JudgeIsIPV6(ip) {
		ip = fmt.Sprintf("%s%s%s", "[", ip, "]")
	}
	sc := []constant.ServerConfig{
		{
			IpAddr: ip,
			Port:   uint64(port),
		},
	}

	cc := constant.ClientConfig{
		NamespaceId:         "public",
		TimeoutMs:           5000,
		NotLoadCacheAtStart: true,
		LogDir:              filepath.Join(global.LogsPath(), "nacos", name),
		CacheDir:            filepath.Join(global.CachePath(), "nacos", name),
		Username:            username,
		Password:            password,
		LogRollingConfig: &constant.ClientLogRollingConfig{
			MaxSize:    8 * 1024 * 1024,
			MaxAge:     7,
			MaxBackups: 5,
			Compress:   true,
			LocalTime:  true,
		},
	}

	// 初始化Nacos链接客户端
	client, _:= clients.NewNamingClient(vo.NacosClientParam{
		ClientConfig:  &cc,
		ServerConfigs: sc,
	})


```
Then execute:
```shell
go run -race main.go
```

you will see
```text
==================
WARNING: DATA RACE
Read at 0x00c000b2b2d8 by goroutine 43:
  github.com/nacos-group/nacos-sdk-go/v2/common/remote/rpc.(*RpcClient).sendHealthCheck()
      rpc_client.go:519 +0x44

Previous write at 0x00c000b2b2d8 by main goroutine:
  github.com/nacos-group/nacos-sdk-go/v2/common/remote/rpc.(*RpcClient).Start()
      rpc_client.go:339 +0x7f9
==================
```

**Expected behavior**

RpcClient should safely synchronize concurrent access to its internal fields during Start() and background goroutines such as sendHealthCheck(). All shared field access should be protected via sync.Mutex or atomic.

**Environment**

SDK version: v2.3.1

[...]

PR #899, opened 2026-07-20
feat: expose auth login errors, TTL-based token refresh, and fix NacosAuthClient data races

### LINKED ISSUE #670: 登录 nacos 时 为啥失败后 会忽略 登录错误？
v2.2.3

v2.2.3

go1.20.3
mac 14.0 (23A344)

### What did you do?

### What did you see instead?
github.com/nacos-group/nacos-sdk-go/v2@v2.2.3/common/nacos_server/nacos_server.go
NewNacosServer 函数
image

cxhello:
## What this does

PR3 of the v3 upgrade roadmap (#879), the Auth layer. Builds on PR1 (#881) and PR2 (#897). Refactors `common/security` to expose login failures, refresh tokens by TTL, and eliminate data races — while keeping the existing `SecurityProxy`/`AuthClient` type names (incremental migration, no breaking rename).

### Changes
1. **Expose login errors (#670).** `SecurityProxy.Login()` now returns an aggregated `error`. New `ClientConfig.FailOnAuthError` (default `false`) makes `NewClient` fail fast when the initial login fails — i.e. **any non-200 response** from the auth endpoint, **or a 200 response carrying no usable token** (both surface as `errors.Is(err, security.ErrLoginFailed)`). Only **transport errors** (server unreachable) stay transient and keep retrying, so a K8s startup-ordering race never aborts construction. Level-graded logging (`Errorf` for login failures, `Warnf` for transient) lets users distinguish "wrong password" from "server unreachable" even without the flag.

   > **Why not just 401/403?** Probed against real servers: Nacos 2.5.2 returns **403** for a wrong password but **500** for an unknown user, while 3.2.0 is the **reverse**, and 3.x's 500 body carries no credential hint. No single input yields 401/403 on both versions, so a 401/403-only rule cannot reliably catch bad credentials. Any non-200 means no token was issued, so all are treated as a login failure.
2. **TTL-based refresh.** `AuthClient` gains `AutoRefresh(ctx)`; each client self-schedules — `NacosAuthClient` refreshes on a token-TTL timer, `RamAuthClient` is a no-op (RAM/STS signing happens per request). Removes `SecurityProxy`'s unconditional 5s polling loop.
3. **Concurrency safety.** `NacosAuthClient`'s refresh state (`lastRefreshTime`/`tokenTtl`/`tokenRefreshWindow`/`serverCfgs`) is now mutex-guarded; `accessToken` stays an `atomic.Value`; no lock is held across the login HTTP IO. Verified with a `-race` concurrent test.

### #807 / #728
Their core was already fixed on this branch by earlier commits (token-refresh loop: f1545e0; endpoint-mode login: #730). This PR adds regression tests and the error-exposure path; closing both on merge.

## Compatibility
- **Default behavior unchanged**: `FailOnAuthError=false` keeps the current "log and continue" behavior for existing users.
- **`AuthClient` interface** gained `AutoRefresh(ctx)` and `Login()`'s return type changed — a breaking change for any external implementer of this interface. v3 is a new module path with no existing external implementations; all in-repo callers are updated.

## Incidental fix (outside the auth layer)

`common/remote/rpc/rpc_client.go` read `rpcClientStatus` non-atomically in `Request()`'s not-connected path (via the value-receiver `getDesc()`, which copies the field), while every other access to that field uses `CompareAndSwapInt32` / `StoreInt32` / `LoadInt32`. When a request runs concurrently with `reconnect()` storing a new status, `-race` flags it — this intermittently failed the `TestIntegrationReconnect` job on the Nacos 2.x runner (~1 in 4 CI runs; not reproducible locally in 5 consecutive runs).

The line dates to 2022 and is the same class as the `currentConnection` data race fixed in #833, which missed this one read. It is unrelated to the auth changes, but it was blocking this PR's CI, so it is fixed here as a single-line change. Happy to split it into its own PR if you'd prefer.

## Non-goals
- No migration of auth to proto/gRPC (login is an HTTP channel by design).
- No changes to RAM/KMS credential-signing internals.
- Config/Naming module issues (#630/#733) are out of scope (separate PRs).

## Testing
- Unit: login success / non-200 classification / malformed-200 rejection (missing, empty or non-string `accessToken`; missing, zero, negative or sub-second `tokenTtl`); deterministic refresh-deadline and clamp tests; `FailOnAuthError` both sides; short-TTL auto-refresh actually fires (covers #807 regression); `UpdateServerList` ‖ `Login` ‖ `GetServerList` under `-race`; `SecurityProxy` aggregation semantics; `NewNacosServer` credential-abort via `httptest` 401.
- Integration (`//go:build integration`, CI Nacos 2.5.2 + 3.x, auth-enabled): wrong password + `FailOnAuthError=true` → `NewConfigClient` returns error; default → constructs a usable client.
- Local: `gofmt`/`goimports`/`go build`/`go vet` clean (one pre-existing unrelated `naming_http` vet warning); `go test ./common/... -race` all pass.

## Related
- Fixes #670
- Refs #807, #728 (regression coverage; closed on merge), #879 (roadmap), #898 (sub-issue)

### DIFF
--- a/clients/config_client/config_client.go
+++ b/clients/config_client/config_client.go
@@ -110,24 +110,29 @@ func NewConfigClientWithRamCredentialProvider(nc nacos_client.INacosClient, prov
 	config.INacosClient = nc
 	clientConfig, err := nc.GetClientConfig()
 	if err != nil {
+		config.cancel()
 		return nil, err
 	}
 	serverConfig, err := nc.GetServerConfig()
 	if err != nil {
+		config.cancel()
 		return nil, err
 	}
 	httpAgent, err := nc.GetHttpAgent()
 	if err != nil {
+		config.cancel()
 		return nil, err
 	}
 
 	if err = initLogger(clientConfig); err != nil {
+		config.cancel()
 		return nil, err
 	}
 	clientConfig.CacheDir = clientConfig.CacheDir + string(os.PathSeparator) + "config"
 	config.configCacheDir = clientConfig.CacheDir
 
 	if config.configProxy, err = NewConfigProxyWithRamCredentialProvider(config.ctx, serverConfig, clientConfig, httpAgent, provider); err != nil {
+		config.cancel()
 		return nil, err
 	}
 
@@ -145,6 +150,7 @@ func NewConfigClientWithRamCredentialProvider(nc nacos_client.INacosClient, prov
 
 	uid, err := uuid.NewV4()
 	if err != nil {
+		config.cancel()
 		return nil, err
 	}
 
--- a/common/constant/client_config_options.go
+++ b/common/constant/client_config_options.go
@@ -252,3 +252,10 @@ func WithClientIP(clientIP string) ClientOption {
 		config.ClientIP = clientIP
 	}
 }
+
+// WithFailOnAuthError ...
+func WithFailOnAuthError(failOnAuthError bool) ClientOption {
+	return func(config *ClientConfig) {
+		config.FailOnAuthError = failOnAuthError
+	}
+}
--- a/common/constant/client_config_options_test.go
+++ b/common/constant/client_config_options_test.go
@@ -93,3 +93,9 @@ func TestNewClientConfigWithOptions(t *testing.T) {
 	assert.Equal(t, config.AccessKey, "accessKey_1")
 	assert.Equal(t, config.SecretKey, "secretKey_1")
 }
+
+func TestWithFailOnAuthError(t *testing.T) {
+	config := ClientConfig{}
+	WithFailOnAuthError(true)(&config)
+	assert.True(t, config.FailOnAuthError)
+}
--- a/common/constant/config.go
+++ b/common/constant/config.go
@@ -62,6 +62,7 @@ type ClientConfig struct {
 	ClusterName          string                   // the address server  clusterName
 	AppConnLabels        map[string]string        // app conn labels
 	ClientIP             string                   // the custom client ip, if not set, will use local ip auto detected
+	FailOnAuthError      bool                     // if true, NewClient returns an error when the initial credential login fails (HTTP 401/403); default false keeps backward-compatible behavior
 }
 
 type ClientLogSamplingConfig struct {
--- a/common/constant/config.go
+++ b/common/constant/config.go
@@ -62,7 +62,7 @@ type ClientConfig struct {
 	ClusterName          string                   // the address server  clusterName
 	AppConnLabels        map[string]string        // app conn labels
 	ClientIP             string                   // the custom client ip, if not set, will use local ip auto detected
-	FailOnAuthError      bool                     // if true, NewClient returns an error when the initial login gets any non-200 response from the auth endpoint (Nacos 2.x/3.x report rejected credentials with an inconsistent mix of 401/403/500); transport errors (server unreachable) stay retryable. Default false keeps backward-compatible behavior
+	FailOnAuthError      bool                     // if true, NewClient returns an error when the initial login gets any non-200 response from the auth endpoint (Nacos 2.x/3.x report rejected credentials with an inconsistent mix of 401/403/500) or a 200 response carrying no usable token; transport errors (server unreachable) stay retryable. Default false keeps backward-compatible behavior
 }
 
 type ClientLogSamplingConfig struct {
--- a/common/constant/config.go
+++ b/common/constant/config.go
@@ -62,7 +62,7 @@ type ClientConfig struct {
 	ClusterName          string                   // the address server  clusterName
 	AppConnLabels        map[string]string        // app conn labels
 	ClientIP             string                   // the custom client ip, if not set, will use local ip auto detected
-	FailOnAuthError      bool                     // if true, NewClient returns an error when the initial credential login fails (HTTP 401/403); default false keeps backward-compatible behavior
+	FailOnAuthError      bool                     // if true, NewClient returns an error when the initial login gets any non-200 response from the auth endpoint (Nacos 2.x/3.x report rejected credentials with an inconsistent mix of 401/403/500); transport errors (server unreachable) stay retryable. Default false keeps backward-compatible behavior
 }
 
 type ClientLogSamplingConfig struct {
--- a/common/nacos_server/nacos_server.go
+++ b/common/nacos_server/nacos_server.go
@@ -18,6 +18,7 @@ package nacos_server
 
 import (
 	"context"
+	stderrors "errors"
 	"io"
 	"math/rand"
 	"net/http"
@@ -91,7 +92,11 @@ func NewNacosServerWithRamCredentialProvider(ctx context.Context, serverList []c
 		ns.initRefreshSrvIfNeed(ctx)
 	}
 
-	ns.securityLogin.Login()
+	if err := ns.securityLogin.Login(); err != nil {
+		if clientCfg.FailOnAuthError && stderrors.Is(err, security.ErrLoginFailed) {
+			return nil, err
+		}
+	}
 	ns.securityLogin.AutoRefresh(ctx)
 	return &ns, nil
 }
--- a/common/nacos_server/nacos_server_test.go
+++ b/common/nacos_server/nacos_server_test.go
@@ -18,6 +18,12 @@ package nacos_server
 
 import (
 	"context"
+	"errors"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
 	"testing"
 
 	"github.com/nacos-group/nacos-sdk-go/v3/common/http_agent"
@@ -184,4 +190,57 @@ func TestNacosServer_UpdateServerListForSecurityLogin(t *testing.T) {
 	client, ok := nacosAuthClient.(*security.NacosAuthClient)
 	assert.True(t, ok)
 	assert.Equal(t, server.GetServerList(), client.GetServerList())
+}
+
+// newUnauthorizedLoginServer starts an httptest server whose login endpoint
+// always returns 401, and returns a ServerConfig pointing at it.
+func newUnauthorizedLoginServer(t *testing.T) (*httptest.Server, constant.ServerConfig) {
+	t.Helper()
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusUnauthorized)
+		_, _ = w.Write([]byte("unknown user!"))
+	}))
+	u, err := url.Parse(ts.URL)
+	assert.NoError(t, err)
+	host, portStr, err := net.SplitHostPort(u.Host)
+	assert.NoError(t, err)
+	port, err := strconv.ParseUint(portStr, 10, 64)
+	assert.NoError(t, err)
+	return ts, constant.ServerConfig{Scheme: "http", IpAddr: host, Port: port, ContextPath: "/nacos"}
+}
+
+func TestNewNacosServer_FailOnAuthError_AbortsOnCredentialError(t *testing.T) {
+	ts, sc := newUnauthorizedLoginServer(t)
+	defer ts.Close()
+
+	clientConfig := constant.ClientConfig{
+		Username:        "wrong",
+		Password:        "wrong",
+		TimeoutMs:       3000,
+		FailOnAuthError: true,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // stop the AutoRefresh goroutine before the httptest server closes
+	_, err := NewNacosServer(ctx, []constant.ServerConfig{sc},
+		clientConfig, &http_agent.HttpAgent{}, 3000, "", nil)
+	assert.Error(t, err)
+	assert.True(t, errors.Is(err, security.ErrLoginFailed))
+}
+
+func TestNewNacosServer_FailOnAuthError_Disabled_DoesNotAbort(t *testing.T) {
+	ts, sc := newUnauthorizedLoginServer(t)
+	defer ts.Close()
+
+	clientConfig := constant.ClientConfig{
+		Username:        "wrong",
+		Password:        "wrong",
+		TimeoutMs:       3000,
+		FailOnAuthError: false,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // stop the AutoRefresh goroutine before the httptest server closes
+	server, err := NewNacosServer(ctx, []constant.ServerConfig{sc},
+		clientConfig, &http_agent.HttpAgent{}, 3000, "", nil)
+	assert.NoError(t, err)
+	assert.NotNil(t, server)
 }
--- a/common/remote/rpc/rpc_client.go
+++ b/common/remote/rpc/rpc_client.go
@@ -616,7 +616,8 @@ func (r *RpcClient) Request(request rpc_request.IRequest, timeoutMills int64) (r
 		conn := r.GetCurrentConnection()
 		if conn == nil || !r.IsRunning() {
 			currentErr = waitReconnect(timeoutMills, &retryTimes, request,
-				errors.Errorf("client not connected, current status:%s", r.rpcClientStatus.getDesc()))
+				errors.Errorf("client not connected, current status:%s",
+					RpcClientStatus(atomic.LoadInt32((*int32)(&r.rpcClientStatus))).getDesc()))
 			continue
 		}
 		response, err := conn.request(request, timeoutMills, r)
--- a/common/security/errors.go
+++ b/common/security/errors.go
@@ -1,0 +1,39 @@
+/*
+ * Copyright 1999-2020 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package security
+
+import (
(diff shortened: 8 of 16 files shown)


cxhello:
@Sunrisea PR3（Auth 层）已就绪，CI 8/8 全绿，麻烦帮忙 review 🙏

这是 roadmap #879 的第三个 PR，基于已 merge 的 PR1(#881)/PR2(#897)。三个核心改动：

1. **暴露登录错误(#670)** — `SecurityProxy.Login()` 返回聚合 error；新增 `ClientConfig.FailOnAuthError`（默认 `false`，向后兼容）。日志分级：凭据错误 `Errorf`、transient `Warnf`，即使不开开关也能区分「密码错」和「网没通」。
2. **TTL 定时刷新** — `AuthClient` 增加 `AutoRefresh(ctx)`，各 client 自调度；删除 `SecurityProxy` 无条件的 5s 轮询。
3. **并发安全** — `NacosAuthClient` 刷新状态加 mutex（`accessToken` 仍 `atomic.Value`），IO 不持锁，`-race` 验证。

**有一处设计决策想请你重点看**（也是我实测踩的坑）：

`FailOnAuthError` 的 fail-fast 判据，我最初设计成「仅 401/403 凭据错误」。但集成测试暴露出 Nacos **不同版本对凭据失败的 HTTP 码不一致**，我起真实容器探针确认：

| 场景 | Nacos 2.5.2 | Nacos 3.2.0 |
|---|---|---|
| 已存在用户 + 错误密码 | **403** | **500** |
| 不存在的用户 | **500** | **403** |

即两个版本各有一种凭据失败被服务端返回成 HTTP 500，且 3.x 的 500 body 是 `"No message available"`，无任何凭据线索。没有单一输入能在两版本都拿到 401/403。

因此我把 `classifyLoginStatus` 改为：**login 接口任何非-200 响应都视为登录失败（wrap `ErrLoginFailed`）**，只有 transport error（连不上服务器）保持 transient 重试。这样：
- FailOnAuthError 在 2.x/3.x 都能可靠 fail-fast（已双版本实测）；
- K8s 启动顺序竞态（Nacos 还没起来）走 transport-error 路径，仍不会误杀 `NewClient`；
- 默认 `false` 时行为不变（登录失败只记日志、继续、AutoRefresh 重试）。

如果你觉得「500 也 fail-fast」过于激进、更希望保守到只认 4xx，我可以调整——但那样 3.x 的错误密码就漏判了，想听下你的意见。

**#807/#728**：核心已被此前 commit（token 刷新 f1545e0 / endpoint 模式 #730）修复，本 PR 补了回归测试与错误暴露路径，merge 后关闭这两个 issue。

`AuthClient` 接口新增了方法、`Login()` 返回值变了——对外部实现者是 breaking，但 v3 是新 module path 无存量外部实现，仓内调用方都已更新。

Sunrisea:
Thanks for putting this together. I reviewed the full Auth-layer change.

Overall, I agree with the direction of this PR:

- `FailOnAuthError` is opt-in, so the default client construction behavior remains backward compatible.
- Moving token refresh scheduling into each `AuthClient` is cleaner than polling every client every 5 seconds.
- The mutex coverage around refresh state and server-list snapshots addresses the reported races without holding a lock across HTTP requests.
- `RamAuthClient.AutoRefresh()` being a no-op is consistent with its per-request credential refresh model.
- The unit and integration test coverage is generally good, and all current CI checks are green.

Regarding the non-200 classification: given the observed differences between Nacos 2.5.2 and 3.2.0, treating any non-200 response from the login endpoint as `ErrLoginFailed` is a reasonable pragmatic choice. Since `FailOnAuthError` defaults to `false`, existing users are unaffected. The trade-off is that users enabling this option will also fail fast on a genuine server-side 5xx response, so this behavior needs to be documented accurately.

Before merging, I think the following issues should be addressed.

### 1. Successful HTTP responses are not validated before enabling refresh

`NacosAuthClient.login()` currently returns `true, nil` for any valid JSON response with HTTP 200, even when `accessToken` is missing.

When `accessToken` exists but `tokenTtl` is missing, has the wrong type, or is non-positive, the type assertion silently produces zero. `AutoRefresh()` then treats the login as successful and resets its timer using `ttl-window`. With both values equal to zero, `timer.Reset(0)` can create a tight login loop and continuously hit the auth endpoint.

A non-string `accessToken` can also be stored in `atomic.Value` and later panic when it is asserted as a string.

Please validate that:

- `accessToken` is a non-empty string;
- `tokenTtl` is present and positive;
- the calculated refresh delay is positive.

An invalid success response should return a non-nil error and use the bounded retry path. A malformed-response unit test should also be added.

### 2. `Login()` and `AutoRefresh()` use different refresh deadlines

`Login()` allows a refresh at:

    lastRefreshTime + tokenTtl - 2*tokenRefreshWindow

while `AutoRefresh()` schedules the next attempt after:

    tokenTtl - tokenRefreshWindow

For a 10-second TTL and a 1-second window, the intended refresh point is 8 seconds, but the automatic refresh actually happens at 9 seconds. This reduces the intended safety margin from two windows to one.

There is also potential timer drift. If `Login()` returns successfully without making an HTTP request because the token is still inside the guard window, `AutoRefresh()` still starts a full `ttl-window` delay from the current time. The next attempt can therefore move close to, or beyond, the actual token expiration time.

Please centralize the refresh deadline calculation and calculate the timer delay from `lastRefreshTime`. Both the guard and scheduler should use the same rule, with non-positive delays clamped safely.

The existing test says the refresh should happen after 8 seconds, but its 12-second timeout also accepts the current 9-second implementation. A deterministic test for the calculated refresh deadline would catch this mismatch more reliably.

### 3. Documentation does not match the implemented classification

`ClientConfig.FailOnAuthError` still says that it applies to HTTP 401/403, and the PR description says that 5xx responses remain transient. The current implementation classifies every non-200 login response as `ErrLoginFailed`.

Please update the configuration comment and PR description to explicitly state that enabling `FailOnAuthError` makes any non-200 login response fail client construction, while transport errors remain retryable.

The `AuthClient` interface change is technically breaking for external implementations, but given that this targets the new v3 module path and the PR already calls it out, I do not consider that a blocker.

My recommendation is **request changes** for items 1 and 2, then approve after the response validation, refresh scheduling, and related tests are updated.

cxhello:
Thanks for the detailed review — all three items were valid and are now addressed. CI is green (8/8).

**Item 1 — validate the success response.** `login()` now requires `accessToken` to be a **non-empty string** and `tokenTtl` to be **present and positive** before it stores anything or arms the refresh state; an invalid 200 returns a non-nil error and goes down the bounded-retry path. Added a table test covering all six malformed shapes (missing / empty / non-string `accessToken`; missing / zero / negative `tokenTtl`), plus an assertion that a rejected response leaves the refresh state unarmed.

Reviewing my own fix, I found one more hole worth calling out: the positivity check ran on the `float64` while the stored value was `int64(ttl)`, so `"tokenTtl": 0.5` passed validation and still landed `tokenTtl = 0` — reintroducing your exact failure mode at a 5s cadence instead of a busy loop. It now validates the **converted** value. I also floored `tokenRefreshWindow` at 1s, since `ttl/10` is 0 for any ttl < 10 and collapsed the deadline onto the expiry instant.

**Item 2 — unified refresh deadline.** The deadline (`lastRefreshTime + tokenTtl - 2*tokenRefreshWindow`) now lives in one place, `refreshDeadlineUnix()`. Both the `login()` guard and the `AutoRefresh()` scheduler derive from it, and the timer delay is computed as `deadline - now`, anchored to `lastRefreshTime` — so a `Login()` that short-circuits on a still-valid token no longer pushes the next attempt out. Non-positive delays are clamped. Deterministic tests pin the exact deadline value, the not-logged-in case, the past-deadline clamp, and a future-deadline delay (5s under the corrected rule vs 9s under the old one). The pre-existing timing test now fires at ~8s, matching the intended two-window margin.

**Item 3 — docs.** Updated `ClientConfig.FailOnAuthError`, the `ErrLoginFailed` sentinel doc, and the PR description. They now state that fail-fast covers any non-200 response **and** a 200 response carrying no usable token, while transport errors stay retryable.

**One decision I'd like your call on:** I classified "200 with no usable token" as `ErrLoginFailed` rather than a plain transient error, on the grounds that no usable token means the login failed. `AutoRefresh` retries on any error either way, so the bounded-retry path you asked for is preserved; the only added effect is that it also fails fast at startup when `FailOnAuthError` is on. If you'd rather that case stay transient and reserve fail-fast for non-200 responses only, it's a one-line change.

**Incidental fix, outside the auth layer.** `TestIntegrationReconnect` started failing the Nacos 2.x job under `-race`. It turned out `rpc_client.go` read `rpcClientStatus` non-atomically in `Request()`'s not-connected path (through the value-receiver `getDesc()`, which copies the field), while every other access uses `CompareAndSwapInt32`/`StoreInt32`/`LoadInt32`. It races with `reconnect()` storing a new status — pre-existing since 2022, same class as the `currentConnection` race fixed in #833, which missed this one read. Roughly 1 in 4 CI runs hit it; I couldn't reproduce it locally in 5 consecutive runs. Fixed as a single line here since it was blocking this PR's CI, but I'm happy to split it into its own PR if you'd prefer that.

Verified against real servers: the full integration suite passes against both Nacos 2.5.2 and 3.2.0 locally (forced with `-count=1`), and CI is green on both.

Sunrisea:
Thanks for the detailed update. I re-reviewed the latest head (`e7c52e6`) and reran the relevant tests.

The three original review items are now addressed correctly:

- Malformed HTTP 200 responses no longer store a token or arm refresh state.
- The login guard and scheduler now share the same deadline anchored to `lastRefreshTime`.
- The configuration comments and PR description match the implemented classification.

On your question: I agree with classifying a 200 response without a usable token as `ErrLoginFailed`. When `FailOnAuthError` is explicitly enabled, returning a client that cannot authenticate would contradict the option's purpose. With the default value unchanged, automatic retries are still preserved for existing users.

I found two remaining issues before approval.

### 1. The malformed-response error can expose the access token

When `accessToken` is a valid string but `tokenTtl` is missing or invalid, the error includes the complete response body:

```go
fmt.Errorf(
    "%w: login response has missing or non-positive tokenTtl: %s",
    ErrLoginFailed,
    string(bytes),
)
```

That error is subsequently logged by `SecurityProxy.Login()` or `AutoRefresh()`, and may also be returned to the caller. This can leak a valid access token into application logs.

Please avoid including the raw response body in these errors, or explicitly redact sensitive fields. It would also be useful to assert that the returned error does not contain the token value.

### 2. Multi-server errors are still order-dependent

`NacosAuthClient.Login()` overwrites `throwable` on every server attempt and returns only the final error.

If one server returns `ErrLoginFailed` and a later server returns a transport error, the transport error hides the login failure. Consequently, `FailOnAuthError=true` may or may not fail client construction depending on the order of the server list.

If all servers fail, the result should preserve any observed `ErrLoginFailed`, either by tracking it separately or aggregating the errors. A successful login from any server should still return success immediately. Please add tests for both server-error orderings.

One minor test gap: the updated description mentions coverage for a sub-second `tokenTtl`, but the malformed-response table does not currently include `tokenTtl: 0.5`. The implementation handles it correctly after conversion, but adding that exact regression case would protect the fix.

The incidental atomic `rpcClientStatus` read looks correct. It is outside the Auth scope, but I do not consider the one-line fix a technical blocker if the maintainers prefer to keep it here.

Local verification passed:

- `go test ./common/security ./common/nacos_server -race -count=1`
- `go test ./common/remote/rpc -race -count=1`
- `go test ./... -short -count=1`

Once the token redaction and multi-server error preservation are addressed, I think this will be ready to approve.

cxhello:
Both issues are fixed in `9b2cb67`, and thanks for catching the token leak — that one was mine, introduced last round when I put the raw body into the error for diagnosability without asking what that body contains at that point in the flow.

**1. Token redaction.** You're right that the `tokenTtl` branch is the dangerous one: `accessToken` has already been validated there, so the body definitely carries a usable token, and the error is logged by `SecurityProxy.Login()` and `AutoRefresh()`. Both malformed-response errors now carry only non-sensitive detail — the offending `tokenTtl` value, never the body — with a comment recording why. Added a test that asserts the returned error does not contain the token value; it failed before the fix with the token printed verbatim.

I left `classifyLoginStatus` still including non-200 bodies: a non-200 response means no token was issued, so it cannot leak one, and those bodies ("unknown user!", the 3.x 500 envelope) are the main diagnostic signal for the cross-version behavior. Happy to redact those too if you'd rather have a blanket rule.

**2. Multi-server error preservation.** `Login()` now tracks the first `ErrLoginFailed` separately from the last error and prefers it, so a credential failure is no longer hidden by a later transport error; a success from any server still returns immediately. Tests cover both orderings — before the fix, `credential-first-then-transport` failed while `transport-first-then-credential` passed, which is exactly the order dependence you described.

**3.** Added the `tokenTtl: 0.5` case to the malformed-response table. As you noted the implementation already handled it after the conversion fix, so this is regression protection.

Verification: full integration suite against real Nacos 2.5.2 and 3.2.0 with `-race -count=1` and `NACOS_CONTAINER_NAME` set so `TestIntegrationReconnect` actually runs — all six tests pass on both versions, no race. (That last flag is what I had been missing locally, which is why the `rpcClientStatus` race only showed up in CI.) CI is green, 8/8.

Sunrisea:
Thanks for addressing the previous review comments. I re-reviewed the latest head (`9b2cb67`). The response validation, refresh deadline calculation, token redaction, and multi-server error preservation are now implemented correctly.

I found two remaining correctness issues and one minor test cleanup item.

### 1. `RamAuthClient` no longer retries a failed provider initialization

Before this PR, `SecurityProxy.AutoRefresh()` called `SecurityProxy.Login()` every five seconds. That also called `RamAuthClient.Login()`, so a matched RAM provider whose `Init()` initially failed would be retried.

After this PR, refresh responsibility is delegated to each `AuthClient`, but `RamAuthClient.AutoRefresh()` is a no-op. The resulting behavior is:

- `RamAuthClient.Login()` selects a provider and calls `Init()`;
- if `Init()` fails, `SecurityProxy.Login()` logs that the transient error "will retry";
- `RamAuthClient.AutoRefresh()` does nothing, so `Init()` is never called again;
- requests continue using an uninitialized provider and generally receive an empty `RamContext`.

This affects built-in providers that need an initialized credential client, as well as custom `RamCredentialProvider` implementations. It also changes the observable lifecycle of custom providers: `Init()` previously ran every five seconds, but now runs only once.

I confirmed the failed-initialization case with a focused test: the provider fails its first `Init()` and would succeed on the second call. After waiting longer than the previous five-second retry interval, `Init()` had still been called only once.

I do not think the old unconditional five-second reinitialization should be restored. A better lifecycle would be:

1. Retry `Init()` every five seconds only when a provider matched but initialization has not succeeded.
2. Stop the retry loop after the first successful initialization.
3. Continue relying on `GetCredentialsForNacosClient()` for per-request credential refresh.
4. Do not expose a partially initialized provider through `GetSecurityInfo()`.

Please add tests covering an initial failure followed by success, no repeated `Init()` after success, and context cancellation while retrying.

### 2. Endpoint mode can leak the server-list refresh goroutine when fail-fast authentication aborts ConfigClient construction

`NewNacosServerWithRamCredentialProvider()` starts `initRefreshSrvIfNeed(ctx)` before the initial authentication login. In endpoint mode, this starts the background server-list refresh goroutine.

With `FailOnAuthError=true`, this PR adds a new return path after that goroutine has started:

```go
if clientCfg.FailOnAuthError && errors.Is(err, security.ErrLoginFailed) {
    return nil, err
}
```

`NewConfigClient()` owns the context passed into this constructor, but its `NewConfigProxy()` error path returns without calling `config.cancel()`. The caller receives no client and therefore cannot stop the goroutine through `CloseClient()`.

The missing ConfigClient cleanup is pre-existing, but the fail-fast authentication path introduced by this PR makes it directly reachable through the new feature. NamingClient already cancels its context on the corresponding construction failure.

Please cancel the ConfigClient context before returning a construction error, or make `NacosServer` cancel its endpoint refresh context when its own constructor fails.

### 3. The non-fail-fast integration test leaves its client running

`TestIntegrationAuth_WrongPasswordDefaultNoFailFast` creates a ConfigClient but never closes it. Its authentication refresh task continues retrying the intentionally incorrect password every five seconds for the remainder of the integration test process.

Please use `require.NoError` / `require.NotNil` and then register `defer client.CloseClient()`.

The new Nacos token refresh mechanism looks good after the latest fixes. My remaining blocker is the RAM initialization retry regression. I also recommend fixing the endpoint construction cleanup in this PR because the new fail-fast path is what exposes it.

cxhello:
All three are fixed in `9f0a2aa`. The RAM one was a real regression on my side — thanks for reproducing it with a focused test.

**1. RAM provider initialization retry.** You diagnosed it exactly: when I moved refresh scheduling into each `AuthClient`, I reasoned that `RamAuthClient` needed no background work because credentials are resolved per request. That was true of the *credentials*, but the five-second `SecurityProxy` poll I deleted had also been retrying `Init()` as a side effect, and I didn't account for that job when removing it.

Implemented along the lifecycle you outlined: `AutoRefresh()` retries `Init()` only while a provider has matched but initialization has not succeeded, stops on the first success, and returns on context cancellation. `Login()` no longer re-runs `Init()` once initialized, so the old unconditional five-second reinitialization is not restored and custom providers don't see a repeating `Init()`. `GetSecurityInfo()` now returns empty unless the provider is both matched and initialized, so a partially initialized provider is never used for signing.

`matchedProvider` and `initialized` are now mutex-guarded: the retry goroutine runs concurrently with per-request credential lookups, so without it `-race` would flag them.

Tests cover initial failure then success (asserting `Init` runs again), no further `Init` after success, context cancellation stopping the retries, and that an uninitialized provider is never consulted for credentials. The retry interval is an injectable struct field rather than a package variable, so shortening it in tests can't race with other tests' goroutines.

**2. Endpoint construction cleanup.** Fixed in `ConfigClient`, matching what `NamingClient` already does. I cancelled on all six construction error paths rather than only the reachable one, since they all sit after the context is created.

**3.** `TestIntegrationAuth_WrongPasswordDefaultNoFailFast` now uses `require` and closes the client.

Verification: full integration suite against real Nacos 2.5.2 and 3.2.0 with `-race -count=1` and `NACOS_CONTAINER_NAME` set so `TestIntegrationReconnect` actually runs — all six tests pass on both, no race. CI is green, 8/8.

Sunrisea:
LGTM

Issue #900, opened 2026-07-23
PR4: Naming — sdk-proto migration, auto-batch multi-instance registration, subscribe dedupe, FuzzyWatch


cxhello:
Sub-task of #879 (PR4 of the 7-PR roadmap). Scope:

1. **Proto migration**: encode/decode all naming requests, responses and `NotifySubscriberRequest` push through nacos-sdk-proto adapters (extends the PR2 dispatch pattern).
2. **Auto-batch registration** (#866, #780): the server keeps exactly one publication per (connection, service) — a single register **replaces** it (verified empirically on Nacos 2.5.2 / 3.1.0 / 3.2.0). The SDK now accumulates registered instances per service and upgrades to `BatchInstanceRequest` when the set grows past one; deregister shrinks the set; reconnect redo replays the full set (fixing the redo-cache overwrite and wrong-removal bugs).
3. **Subscribe dedupe** (#655): repeated `Subscribe` with the same callback no longer multiplies callbacks; `Unsubscribe` with a fresh param struct now actually removes.
4. **FuzzyWatch** (#859, naming side): `INamingClient.FuzzyWatch/CancelFuzzyWatch` with Java-parity five-mode pattern matching, batched init sync, diff sync on reconnect. Config side stays in PR5.
5. **#865**: not reproducible against 2.5.2/3.1.0/3.2.0 with the current code base (root cause was pre-#838 v2 SDK); covered by new integration regression tests instead.

Branch: `feat/v3-naming`, targets `v3.x-dev`.

cxhello:
Both halves of PR4 are now merged into `v3.x-dev`: #901 (proto migration, Java-parity redo, auto-batch registration, Nacos 3.x compat) on 2026-07-31, and #904 (FuzzyWatch) on 2026-08-31. Closing manually since merges into a non-default branch don't trigger auto-close.

Issue #859, opened 2025-11-30
Fuzzy service subscribe and config listen


robocanic:
### Use case(s) - what problem will this feature solve?
Admin ops need to do list-watch all configs and services on nacos.

### Proposed Solution
refer to [java implmentionm](https://nacos.io/docs/latest/manual/user/java-sdk/usage/?spm=5238cd80.19c1ee9e.0.0.7b385810l15key#38-%E9%85%8D%E7%BD%AE%E6%A8%A1%E7%B3%8A%E8%AE%A2%E9%98%85)

### Alternatives Considered
page search all configs periodically

PR #904, opened 2026-07-31
feat(naming): add service FuzzyWatch (#859)

cxhello:
## What is this PR for

This PR carries the FuzzyWatch feature (#859) split out of #901. Following the
round-1 review of this PR, the lifecycle has been **rearchitected from
synchronous WATCH/CANCEL transactions to the Java client's desired-state +
background-reconciliation model** (behavioral reference: Java client 3.2.3,
`NamingFuzzyWatchServiceListHolder` / `NamingFuzzyWatchContext`).

### Architecture

- **Desired state, single sender.** `FuzzyWatch`/`handle.Cancel()` only flip
  local desired state (per-pattern `discard`/`consistentWithServer`/epoch) and
  ring a bell. One reconcile worker goroutine is the sole sender of every
  WATCH/CANCEL RPC: bell + 5s poll + 3-minute full resync (Java cadence).
  Same-pattern lifecycle transitions are serialized by construction; an
  in-flight RPC result is only applied if the pattern's epoch is unchanged, so
  a stale success can never clobber a newer removal or reconnect reset.
- **Failure never rolls back registration.** A WATCH failure leaves the local
  context in place (the server may already hold state, e.g.
  `FUZZY_WATCH_PATTERN_MATCH_COUNT_OVER_LIMIT` is returned *after*
  `syncFuzzyWatcherContext`); transient errors back off 1s and retry, capacity
  errors (50310/50311) surface as typed `FuzzyWatchLoadEvent`s through the
  optional `OnLoadEvent` callback with a 60s suppression window, and the
  pattern keeps retrying on the poll cadence.
- **Watcher-scoped cancellation via handle.** `FuzzyWatch` returns a
  `*FuzzyWatchHandle` (Go `func` values have no reliable identity, so the
  handle is the registration identity). `handle.Cancel()` removes only that
  watcher; the server-side CANCEL is sent by the worker after the last watcher
  is gone, and a watcher registered while the CANCEL is in flight revives the
  pattern (re-checked under lock after the CANCEL confirms — no lost
  registrations, no ghost watches). `handle.MatchedServiceKeys(ctx)` blocks
  until the initial FINISH sync and returns the matched service keys — the
  analogue of Java's `fuzzyWatchWithServiceKeys` future.
- **Reconnect reconciliation.** Disconnect marks every pattern inconsistent;
  the worker re-WATCHes with the real `isInitializing` state and the current
  `receivedGroupKeys`, so the server computes a diff. The previous parallel
  redo cache is deleted — the holder is the single source of desired state,
  which removes the redo-vs-cancel ghost-watch race class entirely.
- **Ability negotiation.** The server's ability table (pushed via
  `SetupAckRequest` after connection setup) is now stored on the connection;
  `FuzzyWatch` against a server without the `fuzzyWatch` ability (2.x) fails
  fast with `naming_client.ErrFuzzyWatchNotSupported` instead of retrying
  forever.
- **Dispatcher hardening.** Per-watcher synced-key tracking (Java
  `FuzzyWatchEventWatcherWrapper` parity): duplicate ADD / phantom DELETE are
  skipped per watcher, and the worker's diff-sync heals any watcher that
  missed events. Callbacks run on a per-pattern drain goroutine under
  `recover()` (a panicking watcher cannot kill the process), and the dispatch
  queue is bounded (1024/pattern) — a dropped notification is re-delivered by
  diff-sync because state is updated before, and independently of, delivery.

### Wire fix found during the rework

`NamingFuzzyWatchRequest` was proto-encoded via protojson, whose field name
`isInitializing` does not match the Java server's Jackson property
`initializing` (derived from the `boolean isInitializing` bean); the server
silently dropped the flag, treated every watch as non-initializing, tagged
initial batches `FUZZY_WATCH_DIFF_SYNC_NOTIFY` and never sent
`FINISH_FUZZY_WATCH_INIT_NOTIFY`. The flag now reaches the server under its
correct wire name (verified against Nacos 3.2.0: initial batches are
INIT-tagged and FINISH arrives). The root cause is fixed upstream in
`nacos-sdk-proto` v1.0.0-beta.9; this PR depends on it and keeps proto
encoding, with a regression test pinning the wire name on both the proto and
legacy-JSON encode paths.

### API surface (v3, first release not yet cut)

- `INamingClient.FuzzyWatch(param *vo.FuzzyWatchParam) (*FuzzyWatchHandle, error)`
  — error only for parameter validation / closed client / missing server
  ability; the RPC outcome never affects registration. `CancelFuzzyWatch` is
  removed; cancellation is `handle.Cancel()`.
- `vo.FuzzyWatchParam{ServiceNamePattern, GroupNamePattern, WatchCallback,
  OnLoadEvent}` (`OnLoadEvent` optional, typed capacity notifications).
- Sentinels re-exported as `naming_client.ErrFuzzyWatchNotSupported` /
  `ErrFuzzyWatchClientClosed` for `errors.Is`.
- On adding methods to `INamingClient`: since v3 is a new module
  (`github.com/nacos-group/nacos-sdk-go/v3`) whose first release has not
  shipped, there are no existing v3 implementers to break; if you'd still
  prefer a separate capability interface, happy to split it out.

## Testing

- **Deterministic unit suite** (scripted fake requester + injected clock, no
  network, `-race`): every scenario from the review — WATCH failure after
  server-side state creation (no rollback, load event, pushes still handled);
  concurrent register/cancel churn convergence; reconnect resync racing a
  cancel (no ghost); transient failure retry to consistency; reconnect before
  FINISH keeps `isInitializing=true`; post-registration match-count limit with
  suppression window; watcher-scoped cancel (CANCEL only after last removal);
  callback panic isolation + backpressure drop healed by diff-sync. Plus
  epoch-race regressions with blocked-in-flight sends, and mutation-verified
  coverage of the revive re-check / cancel gate / suppression / healing paths.
- **Integration** (live servers): Nacos 3.2.0 — 6/6 fuzzy scenarios pass
  `-count=1 -race` twice sequentially, including `MatchedServiceKeys` (exact
  key set) and two-handle watcher-scoped cancel; Nacos 2.5.2 —
  `ErrFuzzyWatchNotSupported` fast-fail verified.
- Full suite: `go build ./... && go vet ./... && gofmt -s -l .` clean;
  `go test ./... -count=1` and `go test ./clients/... -count=1 -race` green.

Refs #859 #900 #879


Sunrisea (changes requested):
Thanks for the thorough implementation and tests. After comparing this PR (`c6f73aa`) with the Nacos Java client 3.2.3 and the Nacos 3.2 server behavior, I think the FuzzyWatch lifecycle still needs to be aligned before merge.

The main concern is that this PR treats WATCH/CANCEL as synchronous transactions, while the Java client uses a local desired-state context plus background reconciliation.

### 1. A failed WATCH response does not mean that the server created no state

The Nacos server calls `syncFuzzyWatcherContext(...)` and publishes `ClientFuzzyWatchEvent` before checking `reachToUpLimit(...)`. Therefore, `FUZZY_WATCH_PATTERN_MATCH_COUNT_OVER_LIMIT` can be returned after the server has already associated the pattern with the connection:

https://github.com/alibaba/nacos/blob/3.2.3/naming/src/main/java/com/alibaba/nacos/naming/remote/rpc/handler/NamingFuzzyWatchRequestHandler.java#L54-L96

The current Go failure path removes the callback, pattern context, and redo state. This can leave:

- the server still watching the pattern;
- no local context capable of processing its pushes;
- no redo entry to reconcile it after reconnect.

The Java client deliberately does not roll back the local watcher on a WATCH RPC failure. It keeps the context inconsistent with the server, retries it in the background, and reports pattern/match-count limits through `FuzzyWatchLoadWatcher`:

https://github.com/alibaba/nacos/blob/3.2.3/client/src/main/java/com/alibaba/nacos/client/naming/cache/NamingFuzzyWatchServiceListHolder.java#L236-L354

Please consider using the same desired-state model instead of assuming that an unsuccessful response proves that no server-side state exists.

### 2. Cancellation should be watcher-scoped

The Java API removes the specified watcher only. It sends a server-side CANCEL only after the last watcher for that pattern has been removed:

https://github.com/alibaba/nacos/blob/3.2.3/client/src/main/java/com/alibaba/nacos/client/naming/NacosNamingService.java#L642-L669  
https://github.com/alibaba/nacos/blob/3.2.3/client/src/main/java/com/alibaba/nacos/client/naming/cache/NamingFuzzyWatchContext.java#L213-L295

This PR instead ignores `WatchCallback` and cancels the whole pattern, including unrelated callbacks registered by other callers. Since Go function values cannot be used as reliable identities, I think the API needs an explicit subscription handle/token rather than changing Java's watcher-scoped semantics into pattern-scoped cancellation.

### 3. Concurrent WATCH/CANCEL operations are not atomic

`FuzzyWatch` adds local state, sends WATCH, and then caches redo state. `CancelFuzzyWatch` sends CANCEL, removes redo state, and then unconditionally removes the whole local pattern. These transitions are not serialized or generation-checked.

A possible interleaving is:

1. CANCEL succeeds on the server and removes the redo entry.
2. Before local `RemovePattern`, another `FuzzyWatch` adds a callback.
3. Its WATCH succeeds and returns `nil`.
4. The older CANCEL resumes and removes the entire local pattern.

The successful watcher is then silently lost. Depending on response ordering, the server and redo cache may still contain the pattern while the local holder does not.

The reconnect redo path has the inverse race as well: an in-flight redo WATCH can re-add a pattern after a successful CANCEL, leaving a permanent ghost watch that is re-registered on every reconnect. I do not think this should be deferred as a known limitation.

Please make same-pattern lifecycle transitions atomic, for example with a per-pattern generation/state check or serialized reconciliation. The Java semantics are a useful reference, although the exact Java locking does not need to be copied.

### 4. Reconnect and capacity reconciliation are incomplete

The Java client:

- marks all contexts inconsistent on disconnect;
- continuously retries reconciliation;
- preserves the real `isInitializing` state;
- performs a full reconciliation every three minutes.

https://github.com/alibaba/nacos/blob/3.2.3/client/src/main/java/com/alibaba/nacos/client/naming/remote/gprc/redo/NamingGrpcRedoService.java#L98-L119

The Go implementation performs only one redo attempt in `OnConnected()` and always sends `isInitializing=false`. A transient redo failure disables the watch for the lifetime of that connection.

This also means a pattern that reaches the server's match-count limit after its initial successful registration can silently stop receiving new matching services. Java detects this during periodic reconciliation and invokes `onServiceReachUpLimit()`. Go currently has neither periodic reconciliation nor a typed overload notification; `sendFuzzyWatch` also discards the response `errorCode`.

### 5. Other API/runtime concerns

- `fuzzyWatchWithServiceKeys()` in Java exposes the initial matched-service snapshot and a completion boundary. Go tracks `initialized` internally but exposes neither the completed snapshot nor a FINISH event, so callers cannot know when the initial list-watch is complete.
- Registering an additional watcher for an existing pattern is local-only in Java, with replay of the known matches. Go sends another WATCH RPC, so a transient RPC failure can reject a watcher that Java would register locally.
- Adding methods to the exported `INamingClient` interface is not purely additive in Go: it breaks all external implementations and mocks at compile time.
- The per-pattern callback queue is unbounded. A slow callback can cause unbounded memory growth, and an unhandled callback panic currently terminates the process. Java uses a bounded NotifyCenter publisher, supports a user executor, and isolates callback failures.

Before merge, I suggest adding deterministic tests for:

1. WATCH error after the server has already created state;
2. concurrent successful `FuzzyWatch` and `CancelFuzzyWatch`;
3. redo WATCH racing with CANCEL;
4. transient redo failure followed by retry;
5. reconnect before initial synchronization finishes;
6. reaching the match-count limit after initial registration;
7. multiple watchers with watcher-scoped cancellation;
8. callback backpressure and panic isolation.

In summary, please use the Java SDK's watcher-scoped, desired-state reconciliation model as the behavioral reference. The Go implementation can use a Go-appropriate handle and generation mechanism, but it should preserve the same observable lifecycle and convergence guarantees.

cxhello:
Thanks for the thorough review — the desired-state analysis was spot on. Rather
than patching the transactional model, this round rearchitects the FuzzyWatch
lifecycle to the Java client's desired-state + background-reconciliation model
(commits `66c4997..5fd0da7`, PR description rewritten to match). Point by
point:

**1. Failed WATCH ≠ no server state — adopted.** Registration is now purely
local (`RegisterWatcher` flips desired state and rings the reconcile worker);
an RPC failure never rolls anything back. Capacity rejections
(50310/50311, surfaced with their `errorCode` preserved via a typed
`FuzzyWatchServerError`) are delivered to the optional
`FuzzyWatchParam.OnLoadEvent` callback with a 60s suppression window — the Go
analogue of `FuzzyWatchLoadWatcher.onServiceReachUpLimit/onPatternOverLimit` —
and the pattern keeps retrying on the poll cadence. Covered by
`TestOverLimitKeepsLocalStateAndNotifiesLoad` (pushes arriving after a failed
WATCH are still processed).

**2. Watcher-scoped cancellation — adopted, via the handle you suggested.**
`FuzzyWatch` now returns a `*FuzzyWatchHandle`; `handle.Cancel()` removes only
that registration, and the worker sends the server-side CANCEL only after the
pattern's last watcher is gone (`TestWatcherScopedCancelKeepsOtherWatcher`,
plus a live-server two-handle integration test). `CancelFuzzyWatch` is
removed. `handle.MatchedServiceKeys(ctx)` exposes the initial-sync boundary
(the `fuzzyWatchWithServiceKeys` future analogue).

**3. Atomic lifecycle transitions — adopted via serialized reconciliation.**
All WATCH/CANCEL RPCs are sent by a single worker goroutine; API calls only
flip local desired state. On top of that, each pattern carries an epoch bumped
on every desired-state change; an in-flight RPC's success is applied only if
the epoch is unchanged, and a confirmed CANCEL re-checks discard-and-empty
under lock before removing the pattern. Both interleavings from your review
have dedicated deterministic tests using a requester that blocks mid-send:
`TestRemovalDuringInFlightWatchStillSendsCancelPromptly`,
`TestResetDuringInFlightWatchStillResendsPromptly`,
`TestReviveDuringCancelInFlightKeepsPatternAndResendsWatch`,
`TestReconnectResyncRacingCancelLeavesNoGhost`. The parallel redo cache is
deleted — the holder is the single source of desired state, so the
redo-vs-cancel ghost watch is structurally impossible now.

**4. Reconnect and capacity reconciliation — adopted.** Disconnect marks all
contexts inconsistent; the worker retries on bell + 5s poll with a 1s backoff
on transient errors, performs a 3-minute full resync, sends the real
`isInitializing` (`TestIsInitializingSurvivesReconnectBeforeFinish`) and the
current `receivedGroupKeys` (`TestResendCarriesReceivedGroupKeysSnapshot`),
and no longer discards the response `errorCode`
(`TestMatchCountLimitAfterInitialSuccess` covers the post-registration
over-limit path with the suppression window).

**5. The rest:**
- *Initial snapshot / completion boundary*: `handle.MatchedServiceKeys(ctx)`.
- *Additional watcher on an existing pattern*: local-only with replay of known
  matches — no extra RPC, so a transient failure can no longer reject it.
- *Unbounded queue / panic*: per-pattern queue bounded (1024), callbacks run
  under `recover()`; a dropped notification is healed by the worker's
  per-watcher diff-sync (`TestBackpressureDropHealsViaDiffSync`,
  `TestCallbackPanicIsIsolated`). On the user-executor point: callbacks
  already run on a dedicated per-pattern goroutine (never on the push-handler
  goroutine), so SDK internals are isolated from user code either way; a
  caller who needs a specific execution context can hand off from the
  callback.
- *`INamingClient` additions*: since v3 is a new module whose first release
  has not shipped, there are no existing v3 implementers to break — but if
  you'd prefer a separate capability interface, happy to split it out.

**Found during the rework — a wire bug worth flagging upstream.** The
integration test for `MatchedServiceKeys` exposed that our proto-encoded
`NamingFuzzyWatchRequest` (protojson field `isInitializing`) never reached the
server's Jackson binding, whose property name is `initializing` (derived from
the `boolean isInitializing` bean); the server silently treated every watch as
non-initializing, tagged initial batches DIFF_SYNC and never sent FINISH
(verified fixed against 3.2.0: INIT-tagged batches + FINISH arrive). The root cause is fixed upstream in
`nacos-sdk-proto` v1.0.0-beta.9; this PR depends on it and keeps proto
encoding, with a regression test pinning the wire name on both encode paths.

All eight deterministic scenarios you listed are in
`clients/naming_client/naming_cache/fuzzy_watch_scenarios_test.go` /
`fuzzy_watch_worker_test.go` (fake requester + injected clock, `-race`, with
mutation-verified coverage of the revive re-check, cancel gate, suppression
window and diff-sync healing), and the live-server integration suite passes
6/6 against 3.2.0 plus the 2.x fast-fail via the new ability negotiation
(`SetupAckRequest` → connection ability table).

Sunrisea:
Thanks for the substantial rework. I re-reviewed the latest head (`5fd0da7`) and confirmed that most of the original concerns have been addressed:

- WATCH failures no longer roll back local desired state.
- Cancellation is watcher-scoped through `FuzzyWatchHandle`.
- WATCH/CANCEL/reconnect transitions are serialized by one reconcile worker and protected by epochs.
- Reconnect, periodic full reconciliation, capacity notifications, initial matched keys, bounded callback queues, and panic isolation are now implemented.
- The `initializing` proto wire-name fix in `nacos-sdk-proto` beta.9 is correct.
- The PR description now accurately documents the new lifecycle and the exported v3 API change.

The original race where concurrent FuzzyWatch/CancelFuzzyWatch could silently remove a successful registration appears to be fixed by the handle-based desired-state design.

However, I found three remaining lifecycle/delivery issues. I reproduced all three with deterministic tests against the latest head.

### 1. [P1] Cancel does not invalidate callbacks already queued by the SDK

`HandleSync` / `HandleChangeNotify` snapshot `*watcherEntry` pointers into `notifyTask.targets`. `RemoveWatcherByID` removes the entry from `ctx.entries`, but `deliver` invokes every snapshotted target without checking whether the registration is still active.

A deterministic sequence is:

1. Block the first callback while the drain goroutine is executing it.
2. Enqueue a second event for the same watcher.
3. Call `handle.Cancel()`.
4. Release the first callback.
5. The second callback still runs after cancellation returned.

A callback already executing cannot be recalled, but SDK-owned work that is still pending in the dispatch queue should not invoke a registration after it has been removed. Please mark the entry inactive/canceled during `RemoveWatcherByID` and re-check that state immediately before callback invocation. A regression test should cover enqueue → cancel → drain.

### 2. [P2] FuzzyWatch racing with CloseClient can return a ghost handle

`RegisterWatcher` checks `stopCh` only before calling `ServerSupportsFuzzyWatch()`. That ability check may block for up to two seconds.

If `CloseClient` / `Shutdown` stops the holder while the ability check is in flight, and the check subsequently returns `true`, `RegisterWatcher` still appends the watcher and returns success. The reconcile worker has already exited, so the returned handle can never establish a server-side WATCH, receive events, or complete `MatchedServiceKeys` except by context timeout.

Please linearize the registration commit with `Shutdown`. The ability check can stay outside the lifecycle lock, but after it returns, registration should acquire the same lock used by `Shutdown`, re-check the closed state, and either return `ErrFuzzyWatchClientClosed` or commit the watcher. A plain second channel/atomic check without shared synchronization still leaves a check-to-commit race.

A deterministic test can block `ServerSupportsFuzzyWatch`, call `Shutdown`, release the ability check, and assert that registration returns `ErrFuzzyWatchClientClosed` without creating pattern state.

### 3. [P2] A panicking callback is still marked as successfully delivered

`safeInvoke` recovers a callback panic, but `deliver` unconditionally updates the watcher's `syncedKeys` afterward. Reconciliation therefore considers the event delivered and will not retry it.

The Java client updates `syncServiceKeys` only after `onEvent` returns normally, so a failed callback remains detectable by the next diff-sync pass.

Please have `safeInvoke` report whether the callback completed successfully and update `syncedKeys` only on success. A regression test should verify that a panicking callback remains unsynchronized and can be retried by reconciliation.

### Validation

The existing suite remains green:

- `go test ./... -count=1`
- `go test -race ./clients/naming_client/naming_cache ./clients/naming_client ./clients/naming_client/naming_grpc ./common/remote/rpc -count=1`
- formatting and `git diff --check`

The three additional deterministic probes above fail on the current head, confirming that the remaining concerns are observable runtime behavior rather than race-detector-only findings.

The overall desired-state architecture is now in good shape, but I recommend keeping the review in Changes Requested until these three lifecycle semantics are addressed.

cxhello:
All three issues are fixed in `9210b0b`, each with the deterministic regression test you described.

### 1. Cancel now invalidates callbacks already queued by the SDK

`watcherEntry` gained a `removed` flag (guarded by `ctx.mu`). `RemoveWatcherByID` marks the entry when it takes it out of `ctx.entries`, and `deliver` re-checks the flag immediately before invoking the callback — on both the change-event and load-event paths. A callback already executing still runs to completion (it cannot be recalled), but any task still pending in the dispatch queue is skipped once cancellation has returned.

Regression test: `TestCancelInvalidatesQueuedCallbacks` blocks the first delivery inside the callback, enqueues a second event, cancels the watcher, releases the first delivery, and asserts the queued second task never invokes the removed watcher.

### 2. Registration commit is linearized with Shutdown

`Shutdown` now closes `stopCh` while holding `h.mu`, and `RegisterWatcher` re-checks the closed state under that same lock after the ability check returns, before creating any pattern state. A registration racing `CloseClient` therefore either commits strictly before the close (indistinguishable from registering just before close, as you noted) or fails with `ErrFuzzyWatchClientClosed`. The ability check itself stays outside the lock, and the pre-check before it remains as a fast-fail only.

Regression test: `TestShutdownDuringAbilityCheckRejectsRegistration` blocks `ServerSupportsFuzzyWatch`, calls `Shutdown`, releases the ability check, and asserts the registration returns `ErrFuzzyWatchClientClosed` with no pattern state created.

### 3. A panicking callback is no longer marked delivered

`safeInvoke` now reports whether the callback completed normally, and `deliver` advances `syncedKeys` only on normal completion — matching the Java client, where `syncServiceKeys` is updated only after `onEvent` returns. A panicked delivery leaves the watcher lagging `receivedGroupKeys`, so the reconcile worker's next diff-sync pass re-delivers the event.

Regression test: `TestPanickingCallbackIsRetriedByDiffSync` panics on the first delivery and asserts the worker re-delivers the event via `DIFF_SYNC` and only then marks the key synchronized.

### Validation

- `go test ./... -count=1` — green
- `go test -race ./clients/naming_client/naming_cache ./clients/naming_client ./clients/naming_client/naming_grpc ./common/remote/rpc -count=1` — green
- All 6 fuzzy watch integration tests pass against live Nacos 3.2.0 (and the 2.5.2 fast-fail case)
- `gofmt` / `git diff --check` clean

Sunrisea:
Thanks for the latest fixes. I re-reviewed the current head (`9210b0b`) and confirmed that the three lifecycle issues from my previous review have been addressed.

After reassessing the remaining concerns with a pragmatic merge bar, I only consider one issue blocking. I do not think this PR needs to add per-watcher executors, a strict `Shutdown` join barrier, a per-pattern retry scheduler, or richer error/ability state APIs before merge.

### [P1] `syncVersion` can be stamped current while callback deliveries are still in flight

`syncWatchersLocked` currently advances `entry.syncVersion` whenever the computed diff is empty:

```go
if len(diff) == 0 {
    entry.syncVersion = c.syncVersion
}
```

However, while `c.draining` is true, `entry.syncedKeys` is only an intermediate delivery state. An in-flight or queued callback may still update it after `syncWatchersLocked` returns.

One deterministic sequence is:

1. Block an ADD callback before it updates `syncedKeys`.
2. Fill the bounded queue with another ADD.
3. Apply DELETEs for both keys; `receivedGroupKeys` returns to empty, but the DELETE tasks are dropped because the queue is full.
4. Reconciliation temporarily sees both `receivedGroupKeys` and `syncedKeys` as empty and stamps the watcher with the current `syncVersion`.
5. Release the queued ADD callbacks; they add both keys to `syncedKeys`.
6. Future reconciliation skips the watcher because its version is already current, so the dropped DELETEs are never replayed.

This breaks the documented guarantee that queue drops and failed callback deliveries are healed by diff-sync.

The minimum fix appears sufficient:

```go
if len(diff) == 0 && !c.draining {
    entry.syncVersion = c.syncVersion
}
```

Please also add one deterministic regression test covering a blocked callback, a full queue, and an ADD/DELETE reversal. Once this is fixed and the focused test passes, I think the PR is ready to approve from the lifecycle/design perspective.

cxhello:
Fixed in `c14e702`.

The stamp now requires the context to be at rest: `if len(diff) == 0 && !c.draining`. While a drain is in flight, `syncedKeys` is only an intermediate delivery state, so an empty diff observed mid-drain proves nothing; the watcher keeps its lagging `syncVersion` and the next pass (bell from `onDrop`, or the periodic tick) recomputes the diff after deliveries have settled.

The regression test `TestQueueDropsHealEvenWhenReconcileRunsMidDrain` reproduces your exact sequence deterministically: an ADD blocked inside the callback (before `syncedKeys` is updated), a second ADD filling the limit-1 queue, two DELETEs reverting `receivedGroupKeys` to empty with their tasks dropped, then a reconcile pass run mid-drain. It asserts that after the drain completes, subsequent passes replay both dropped DELETEs as `FUZZY_WATCH_DIFF_SYNC_NOTIFY` events and `syncedKeys` converges back to empty. The test fails on the previous head (the DELETEs are never replayed) and passes with the fix, including under `-race -count=5`.

One note on the test shape: with a limit-1 queue, a replayed diff task can itself be dropped, so the test drives reconcile passes in a loop the same way the bell-driven worker retries after `onDrop` — single-pass convergence is not part of the guarantee being tested.

CI is green on the new head (unit tests on Go 1.22/1.23/1.24, integration tests against Nacos 2.x and 3.x).

Issue #818, opened 2025-05-06
开启 AsyncUpdateService  之后，存在数据竞争问题


helays:
### What version of nacos-sdk-go are you using?
v2.2.9

### What version of nacos-sever are you using?
v2.5.1

### What version of Go are you using (`go version`)?
go version go1.24.0 linux/amd64

### What operating system (Linux, Windows, …) and version?
linux
### What did you do?
If possible, provide a recipe for reproducing the error.
```
==================
WARNING: DATA RACE
Write at 0x00c00067c0f0 by goroutine 1266:
  github.com/nacos-group/nacos-sdk-go/v2/clients/naming_client/naming_cache.instanceSorter.Swap()
      /home/itestor/Go/pkg/mod/github.com/nacos-group/nacos-sdk-go/v2@v2.2.9/clients/naming_client/naming_cache/service_info_holder.go:172 +0x1c4
  github.com/nacos-group/nacos-sdk-go/v2/clients/naming_client/naming_cache.(*instanceSorter).Swap()
      <autogenerated>:1 +0x29
  sort.insertionSort()
      /usr/local/go/go-1.24.0/src/sort/zsortinterface.go:13 +0x81
  sort.pdqsort()
      /usr/local/go/go-1.24.0/src/sort/zsortinterface.go:73 +0x391
  sort.Sort()
      /usr/local/go/go-1.24.0/src/sort/sort.go:54 +0x70
  github.com/nacos-group/nacos-sdk-go/v2/clients/naming_client/naming_cache.sortInstance()
      /home/itestor/Go/pkg/mod/github.com/nacos-group/nacos-sdk-go/v2@v2.2.9/clients/naming_client/naming_cache/service_info_holder.go:190 +0x1bb
  github.com/nacos-group/nacos-sdk-go/v2/clients/naming_client/naming_cache.isServiceInstanceChanged()
      /home/itestor/Go/pkg/mod/github.com/nacos-group/nacos-sdk-go/v2@v2.2.9/clients/naming_client/naming_cache/service_info_holder.go:161 +0x198
  github.com/nacos-group/nacos-sdk-go/v2/clients/naming_client/naming_cache.checkInstanceChanged()
      /home/itestor/Go/pkg/mod/github.com/nacos-group/nacos-sdk-go/v2@v2.2.9/clients/naming_client/naming_cache/service_info_holder.go:140 +0x6db
[...]

PR #906, opened 2026-08-03
fix(naming): copy hosts before sorting in service change detection (#818)

HackenPai:
## What is this PR for

Fixes a data race in `isServiceInstanceChanged` (`clients/naming_client/naming_cache/service_info_holder.go`), where the "old" instance list was sorted in place before being compared against the new one. This race is reliably triggered under `go test -race` when `AsyncUpdateService` is enabled and concurrent `SelectInstances`/`GetServiceInfo` calls are in flight.

## The bug

`isServiceInstanceChanged` compares the previous and new instance lists for a service by sorting copies of both and comparing element-by-element. The "new" side was already copied before sorting:

```go
newInstance := make([]model.Instance, len(newService.Hosts))
copy(newInstance, newService.Hosts)
```

But the "old" side was not — it sorted `oldService.Hosts` directly:

```go
oldInstance := oldService.Hosts  // shares backing array with stored value
sortInstance(oldInstance)         // in-place sort races with readers
```

`oldService` is the `model.Service` value currently stored in `ServiceInfoMap`, and `oldService.Hosts` shares its backing array with that stored value — and with every snapshot previously handed out by `GetServiceInfo`. Sorting `oldInstance` in place therefore mutates memory that a concurrent reader (e.g. another goroutine holding an earlier `GetServiceInfo` result and iterating it in `selectInstances`) may be reading at the same time, which is a classic data race.

Race trace (from #818):

```
WARNING: DATA RACE
Write at ... by goroutine ...:
  instanceSorter.Swap()
  sort.Sort()
  sortInstance()
  isServiceInstanceChanged()
  checkInstanceChanged()
  (*ServiceInfoHolder).ProcessService()
  (*ServiceInfoUpdater).updateServiceNow()
  ...asyncUpdateService...

Previous read at ... by goroutine ...:
  (*NamingClient).selectInstances()
  (*NamingClient).SelectInstances()
```

## The fix

Copy `oldService.Hosts` into a fresh slice before sorting it, mirroring the existing `newInstance` copy:

```go
oldInstance := make([]model.Instance, len(oldService.Hosts))
copy(oldInstance, oldService.Hosts)
```

This is the same fix that was merged into `v3.x-dev` via #903; this PR backports it to `master` (v2.x) so that v2 users are no longer exposed to the race.

## Compatibility notes

- No public API changes.
- No behavioral change for single-threaded callers — the comparison result is identical; only the memory touched during the sort is different.

## Testing

- Unit: `go test ./clients/naming_client/naming_cache/ -count=1 -race` green — existing `TestServiceInfoHolder_isServiceInstanceChanged` and `TestHostReactor_isServiceInstanceChangedWithUnOrdered` pass; added `TestIsServiceInstanceChanged_NoDataRace` which reliably triggers the race detector on the old code (verified by temporarily reverting the fix) and passes cleanly with it.
- Verified: under `-race`, the new test catches the exact `instanceSorter.Swap` vs `selectInstances` race reported in #818 when run against the pre-fix code, and reports zero races after the fix.

Refs #818 #903


HackenPai:
@Sunrisea please review it, thanks

Sunrisea:
lgtm

PR #909, opened 2026-08-20
fix #908: use an atomic load for rpcClientStatus in Request

### LINKED ISSUE #908: Data race on RpcClient.rpcClientStatus in Request()

master (`4ba9445`), also present in `v2.3.5`.

`nacos/nacos-server:v2.1.2`

`go1.26.5`

Linux/amd64 (GitHub Actions `ubuntu-latest`)

### What did you do?

Ran `go test -race` while the server was not yet accepting connections, so `Request()` takes its `conn == nil || !r.IsRunning()` branch.

### What did you expect to see?

No race. `rpcClientStatus` is accessed atomically at 13 of its 14 sites.

### What did you see instead?

The 14th site, `rpc_client.go:619`, reads the field with a plain load:

```go
errors.Errorf("client not connected, current status:%s", r.rpcClientStatus.getDesc())
```

It races with the CAS at the tail of the same function (`rpc_client.go:647`) when two goroutines are inside `Request()` — here, a caller registering an instance and the SDK's own `redoRegisterEachService` after a reconnect:

```
WARNING: DATA RACE
Write at 0x00c000348108 by goroutine 18:
  sync/atomic.CompareAndSwapInt32()
  naming_grpc.(*NamingGrpcProxy).requestToServer()   naming_grpc_proxy.go:88
  naming_grpc.(*NamingGrpcProxy).BatchRegisterInstance()

Previous read at 0x00c000348108 by goroutine 23:
  rpc.(*RpcClient).Request()                         rpc_client.go:619
  naming_grpc.(*NamingGrpcProxy).requestToServer()   naming_grpc_proxy.go:88
  naming_grpc.(*ConnectionEventListener).redoRegisterEachService()
```

Separate from #833/#905, which fixed `currentConnection`.

Full log: https://github.com/gogf/gf/actions/runs/32361418299/job/96401598053

lingcoder:
closes #908

## What

`Request()` reads `rpcClientStatus` without an atomic load while a concurrent `Request()` writes it:

- read — `rpc_client.go:619`, formatting the `client not connected, current status:%s` message
- write — `rpc_client.go:647`, `atomic.CompareAndSwapInt32(..., RUNNING, UNHEALTHY)`

The field is accessed atomically at every other site (13 of 14). `:619` was the only plain one. It is reachable only when the client cannot reach the server, so it surfaces as an intermittent `-race` failure rather than a consistent one.

Separate from #833/#905, which fixed `currentConnection`.

## Change

Add `getStatus()` — the atomic counterpart of the existing `isShutdown`/`IsRunning`/`IsInitialized` helpers — and use it at `:619`. Those three helpers now go through it as well, so `rpcClientStatus` has a single read path and cannot drift again.

## Verification

- `go build ./...`, `go vet ./common/remote/rpc/`, `gofmt -l` clean.
- `go test -race -count=1 ./common/remote/rpc/...` passes.
- New `TestClientStatus_ConcurrentReadWrite` mirrors `TestCurrentConnection_ConcurrentReadWrite` from the `currentConnection` fix. It reports `WARNING: DATA RACE` when `getStatus()` is reverted to a plain field read, and passes with the fix in place — confirmed locally both ways.


Sunrisea:
LGTM

PR #911, opened 2026-08-25
fix: always subscribe services loaded from local disk cache

### LINKED ISSUE #15457: nacos-golang-sdk版本维护问题咨询
目前release的golang sdk只有2.3.5
发现对于naming 模块的客户端缓存策略存在缺陷，默认初始化的情况下，如果有本地缓存，则不会addListener。
但是java sdk v3.1.2就不是这样的特性了

neoLsH:
## Background

Fixes alibaba/nacos#15457 (naming module client cache strategy differs from the Java SDK): when a service's info is loaded from the local disk cache at startup, the Go SDK returns it from `GetService` / `SelectInstances` / `SelectOneHealthyInstance` / `SelectAllInstances` **without ever adding a server-side subscription**, so instance changes are not pushed until an explicit `Subscribe` call. The Java SDK v3 (`tryToSubscribe`) always subscribes in this case and falls back to the local cache only when the subscribe request fails.

## Changes

- `NamingClient.getServiceInfoWithSubscribe` ports the Java `tryToSubscribe` semantics:
  - cache miss → subscribe
  - cache hit + already subscribed → return cache
  - cache hit + not subscribed → subscribe; **fall back to the local cache** when the subscribe request fails
- The four read APIs above now go through this helper instead of returning cache blindly
- `INamingProxy` gains `IsSubscribed`:
  - delegate forwards to the gRPC proxy's existing `IsSubscribed`
  - http proxy always reports `false` (it keeps no push subscription)
  - gomock `MockINamingProxy` updated accordingly

## Tests

4 new unit tests cover all paths (no live server needed):

```
go test ./clients/naming_client/ -run GetServiceInfoWithSubscribe -v
--- PASS: TestNamingClient_GetServiceInfoWithSubscribe_NotCached
--- PASS: TestNamingClient_GetServiceInfoWithSubscribe_CachedAndSubscribed
--- PASS: TestNamingClient_GetServiceInfoWithSubscribe_CachedButNotSubscribed
--- PASS: TestNamingClient_GetServiceInfoWithSubscribe_SubscribeFailsFallbackToCache
```

Also verified: `go build ./...`, `go vet ./clients/...` (no new warnings), `gofmt` clean, and the full `go test ./...` suite passes.


KomachiSion:
[P1] Do not treat a cached redo intent as a successful subscription

`NamingGrpcProxy.IsSubscribed` currently delegates to `eventListener.IsSubscriberCached`, but `Subscribe` adds that redo entry before sending the RPC and leaves it there when the request fails.

As a result, the first read falls back to the local cache as intended, but every subsequent read sees `IsSubscribed == true` and no longer retries the subscription. The cached service may therefore remain stale indefinitely.

This differs from the Java SDK: it keeps redo intent and confirmed subscription state separately. `cacheSubscriberForRedo` creates an entry with `registered=false`, `subscriberRegistered` changes it to `true` only after a successful response, and `isSubscribed` checks that confirmed state. Its scheduled redo task also keeps retrying while `registered=false`.

Please track confirmed subscription state separately and add a regression test covering: cached service → first subscribe fails → second read retries the subscription.

neoLsH:
Thanks for the detailed review — confirmed against the Java SDK (`NamingGrpcRedoService`: `cacheSubscriberForRedo` creates the entry with `registered=false`, `subscriberRegistered` flips it to `true` only after a successful response, `isSubscriberRegistered` checks that confirmed state, and the redo task keeps retrying while unconfirmed).

Fixed in 496e05d:

1. The redo cache now stores a `subscriberRedoEntry` holding an atomic `registered` flag instead of `struct{}{}`, separating the redo intent from the confirmed subscription (aligned with `SubscriberRedoData`).
2. `Subscribe` confirms the entry (`SubscriberRegistered`) only after the request succeeds; on failure the entry stays cached but unconfirmed, so reconnect redo (`OnConnected -> redoSubscribe`) still covers it.
3. `IsSubscribed` now delegates to `IsSubscriberRegistered` — a cached-but-unconfirmed entry is no longer reported as subscribed, so the next read retries.

Added the requested regression test (`TestNamingClient_GetServiceInfoWithSubscribe_FailedSubscribeIsRetried`) covering exactly that sequence: cached service -> first subscribe fails and falls back to the cache -> the next read retries the subscription -> once confirmed, subsequent reads do not subscribe again. `TestSubscriberRedoRegisteredState` additionally covers the entry state machine (cache -> unconfirmed, confirm -> registered, re-cache preserves state, remove -> reset).

Full test suite passes (`go test ./...`).

KomachiSion:
Thanks — `496e05d` fixes the original failed-subscribe retry issue. One connection-lifecycle gap remains.

`OnDisConnect()` is still empty, so subscriptions confirmed on the old gRPC connection remain `registered=true` after that connection is lost. During reconnection, `CacheSubscriberForRedo()` preserves this state. If the redo subscribe request fails, `IsSubscribed()` will still return `true`, preventing subsequent reads from retrying and potentially leaving the cache stale.

The Java SDK resets every subscriber’s `registered` flag to `false` on disconnect before redo. Please do the same here and add a regression test covering:

`confirmed subscription → disconnect → unconfirmed → failed redo remains unconfirmed`

neoLsH:
Thanks — fixed in `3c61a49`: `OnDisConnect()` now resets the confirmed state of every cached subscriber redo entry (`SubscriberOnDisconnect`), aligning with `NamingGrpcRedoService.onDisConnect`. The redo intent itself is preserved, so reconnect redo still covers it; if the redo fails the entry stays unconfirmed and subsequent reads retry.

Added `TestSubscriberRedoOnDisconnect` covering the requested sequence: confirmed subscription → disconnect → unconfirmed (redo intent kept) → failed redo remains unconfirmed → eventual redo success confirms again.

KomachiSion (changes requested):
[P1] Confirm the subscription only after a successful response

`requestToServer` returning a nil error does not necessarily mean that the operation succeeded. `RpcClient.Request` logs a non-`ErrorResponse` with `IsSuccess() == false` but still returns `(response, nil)`.

Here, `SubscriberRegistered` is called before checking `response.IsSuccess()` or validating the response type. Therefore, a failed `SubscribeServiceResponse` can be marked as confirmed: `getServiceInfoWithSubscribe` will return the failed response's service info instead of falling back to the local cache, and subsequent reads will skip retries because `IsSubscribed()` returns `true`.

The Java SDK checks the response result code and throws before calling `subscriberRegistered`.

Please validate that the response is non-nil, has the expected type, and is successful before marking the redo entry as registered. Otherwise, return an error so the cached-service path can fall back and retry later. Please also add a regression test covering:

`SubscribeServiceResponse{success=false} → remains unconfirmed → next read retries`

neoLsH:
Thanks for catching this — confirmed: `RpcClient.Request` only logs non-`ErrorResponse` failures and still returns `(response, nil)`, so a failed `SubscribeServiceResponse` would have been marked as confirmed.

Fixed in 1af220e:

- `Subscribe` now validates the response before confirming: a comma-ok type assertion (an unexpected response type surfaces as an error instead of a panic — the previous code had a bare assertion) followed by an `IsSuccess()` check; `SubscriberRegistered` is called only after both pass, aligning with the Java SDK where a non-success response throws in `requestToServer` before `subscriberRegistered` is reached.
- On failure an error is returned, so the cached-service path in `getServiceInfoWithSubscribe` falls back to the local cache and later reads retry the subscription.
- Added a `requestToServerFn` hook so tests can stub the rpc round-trip, and regression tests covering exactly the requested sequence: `SubscribeServiceResponse{success=false}` → remains unconfirmed (redo intent kept) → next read retries and confirms; plus an unexpected response type → error without panic and no confirmation.

One scope note: the Go rpc layer contract differs from the Java SDK — Java's `requestToServer` translates failed responses, 403 re-login and type checks centrally at the rpc boundary, while Go's `RpcClient.Request` keeps the error dual-track (error = transport failure / retries exhausted), so per-proxy validation is the established idiom here (e.g. `RegisterInstance` returns `response.IsSuccess()`). Centralizing the validation in the rpc layer would change the semantics of all proxy methods, which looks out of scope for this PR — happy to do it as a follow-up if you prefer.

Verified locally: `gofmt -d -s`, `goimports`, `go vet`, `go build ./...` and `go test -race ./...` (19 packages, no races).

Issue #913, opened 2026-09-01
PR5: Config — sdk-proto migration, listen state rebuild, cancel notification, FuzzyWatch


cxhello:
Sub-task of #879 (PR5 of the 7-PR roadmap). Split into two PRs like PR4.

**PR5a (`feat/v3-config`):**

1. **Proto migration**: encode/decode config requests, responses and the `ConfigChangeNotifyRequest` push through nacos-sdk-proto adapters (extends the PR2 dispatch pattern).
2. **Listen state rebuild**: pointer-based cache entries with per-entry locking and listener lists — a second `ListenConfig` on the same key no longer silently drops its callback (Java parity: multiple listeners per config), and the copy-mutate-set races of the by-value cache map go away structurally.
3. **#629**: `CancelListenConfig` now tells the server — cancelled entries are marked discarded and the sync loop sends `ConfigBatchListenRequest{listen=false}` before removing them (probe-verified on 2.5.2/3.2.0 that this stops pushes). Also fixes the corner where cancelling the only listened config sent nothing at all.
4. **#694**: not reproducible on 2.5.2/3.2.0 with the current code base — the connection-event listener synced from master already re-syncs after reconnect (verified by real `docker restart` probes on both versions). Covered by a new restart integration regression test.
5. **#727**: already fixed on the current code base — the server rejects a stale `casMd5` with resultCode 500 on both 2.5.2/3.2.0 and `PublishConfig` propagates it as `(false, error)` (probe-verified). Covered by a new CAS regression test.

**PR5b (`feat/v3-config-fuzzy`):** config-side FuzzyWatch (#859), reusing the desired-state reconcile architecture landed in #904 (single reconcile worker, per-pattern epoch guards, bounded dispatch queue, diff-sync self-healing).

Branch: `feat/v3-config`, targets `v3.x-dev`.

Issue #727, opened 2024-03-02
PublishConfig CasMd5检测不通过发布失败时，无err，并且返回成功；导致应用程序无法判断发布到底是否真的成功


zhangms:
### What version of nacos-sdk-go are you using?
v2.2.5

### What version of nacos-sever are you using?
v2.3.0

### What version of Go are you using (`go version`)?
go 1.21

### What operating system (Linux, Windows, …) and version?
Linux

### What did you do?
func (n *configClientNacos) PublishCas(dataId, groupId, content, casMd5 string) error {
	success, err := n.client.PublishConfig(vo.ConfigParam{
		DataId:  dataId,
		Group:   groupId,
		Content: strings.TrimSpace(content),
		CasMd5:  casMd5,
	})
	if err != nil || !success {
		return errors.Join(errors.New("publish config not success"), err)
	}
	return nil
}

### What did you expect to see?
当参数casMd5 与 nacos-server 上配置的md5不一致时，返回失败或者错误

### What did you see instead?
当参数casMd5 与 nacos-server 上配置的md5不一致时，nacos-server实际未发布成功，但sdk返回了正确

Issue #655, opened 2023-09-06
Repeated subscriptions result in multiple callbacks


binbin0325:
### What version of nacos-sdk-go are you using?
v2.2.3
### What version of nacos-sever are you using?
v2.x
### What version of Go are you using (`go version`)?
v1.18
### What operating system (Linux, Windows, …) and version?
linux
### What did you do?
If possible, provide a recipe for reproducing the error.
![image](https://github.com/nacos-group/nacos-sdk-go/assets/28588342/8547cfb3-0977-43ad-9825-714c8dc0da2c)

![image](https://github.com/nacos-group/nacos-sdk-go/assets/28588342/6bc14ad4-5c10-4cdc-b078-a54b1e4252a8)

cxhello:
In the v3 work we hit this issue directly: Go func values aren't reliably comparable, so dedupe-by-callback-identity (what Java does with listener object identity) can't be implemented soundly — we prototyped code-pointer dedup in #901's early revisions and withdrew it because closures created at the same call site share a code pointer while being semantically distinct.

Instead, #904 (merged) established a handle-based pattern for FuzzyWatch: `FuzzyWatch` returns a `*FuzzyWatchHandle`, and `handle.Cancel()` removes exactly that watcher — no identity guessing. We'd like to propose the same shape for `Subscribe` in v3: return a subscription handle whose `Cancel` removes exactly that callback, making repeated subscriptions well-defined (each gets its own handle) instead of silently stacking. WDYT?

PR #914, opened 2026-09-01
feat(config): migrate config to sdk-proto and rebuild the listen state machine

### LINKED ISSUE #629: cancelling listening is invalid when there is only one configuration
v2.2.2
v2.x
1.18
linux
### What did you do?

cxhello:
## What is this PR for

PR5a of the v3 roadmap (#879, sub-issue #913): migrates the Config module's wire types to nacos-sdk-proto and structurally rebuilds the config listen state machine. Config-side FuzzyWatch (#859) follows separately as PR5b on top of this branch.

### Proto migration (same adapter pattern as PR2/PR4)

- Requests (`ConfigQueryRequest`, `ConfigPublishRequest`, `ConfigRemoveRequest`, `ConfigBatchListenRequest`) implement `ProtoMessage()` and encode through `PayloadCodec`; a wire-parity test pins the protojson key set against the legacy JSON body (guarding the `json_name` defect class found in PR4).
- Responses (`ConfigQueryResponse`, `ConfigPublishResponse`, `ConfigRemoveResponse`, `ConfigChangeBatchListenResponse`) and the `ConfigChangeNotifyRequest` server push decode through `proto_dispatch` adapters back into the legacy structs, so downstream code is unchanged. `success` is derived from `resultCode` (proto has no such field), matching the PR2 rule.
- The legacy-JSON fallback path keeps permanent test coverage via a non-migratable stub request type (it previously piggybacked on `ConfigQueryRequest`, which this PR migrates).
- Note: legacy `ConfigQueryResponse.Tag` is a dead `bool` field (proto's `tag` is a string); the adapter intentionally does not map it.

### Listen state machine rebuild (Java `ClientWorker`/`CacheData` parity)

The by-value `cacheData` copies stored in `cache.ConcurrentMap` (copy-mutate-set races, single silently-overwritten listener) are replaced by a pointer-based `configCacheHolder` with per-entry locking:

- **Multiple listeners per config.** `cacheData.listeners` is a list of wraps, each with its own `lastCallMd5` watermark. Delivery snapshots state under the entry lock, releases it, then per listener: filter-chain decrypt → `recover`-wrapped callback → watermark advances only on normal return. A panicking listener is replayed next round and never blocks its siblings; a filter-chain error skips the round without advancing any watermark.
- **Cancel actually reaches the server (#629).** `CancelListenConfig` marks the entry `discard` and rings the bell; the single executor goroutine sends `ConfigBatchListenRequest{listen=false}` batches (grouped by taskId, before the listen batches) and only removes an entry after the server confirmed — re-checking under lock that no listener re-registered while the cancel was in flight. This also fixes the corner where cancelling the only listened config sent nothing at all.
- **Listen/Close linearization.** `ListenConfig` re-checks `isClosed` under the client mutex after committing its listener and rolls back if a concurrent `CloseClient` won the race; the closed path returns the exported sentinel `ErrConfigClientClosed`. The bell send selects against the client context so post-close notifications cannot leak goroutines.
- The executor keeps the existing cadence (bell + 5s poll + 3min full resync) and the taskId sharding (3000 configs per rpc client). Reconnect handling is unchanged: disconnect marks entries out-of-sync, reconnect rings the bell.

## Issues

- **Fixes #629** — evidence chain: executor unit tests assert the exact `listen=false` request stream; server behavior (push stops after `listen=false`) was probe-verified against Nacos 2.5.2 and 3.2.0; an integration test pins the user-visible contract (no callbacks after cancel). Note the integration test alone cannot prove server-side stop — cancel clears local listeners synchronously — hence the unit + probe layers.
- **#694** — not reproducible on this code base: real `docker restart` probes against 2.5.2 and 3.2.0 show listen callbacks resume after a server restart (the connection-event listener synced from master already re-syncs). A restart integration regression test pins it.
- **#727** — already fixed on `v3.x-dev`: both server versions reject a stale `casMd5` with resultCode 500 and `PublishConfig` propagates `(false, error)`. A dual-version CAS integration regression test pins it. No typed sentinel is introduced: the only discriminator is the server's message string, which is too fragile to key on.

## Behavior changes vs v2 (please review deliberately)

1. **Repeated `ListenConfig` on the same key now appends listeners** (Java parity). v2 silently dropped every callback after the first. Code that called `ListenConfig` repeatedly as an idempotent re-registration (e.g. in reconnect loops) will now accumulate listeners and receive duplicate callbacks — this is the most user-visible change in the PR.
2. **`CancelListenConfig` now notifies the server** and removes *all* listeners of the key (same key-level granularity as v2's local removal). Watcher-scoped cancellation (handle-based, like `FuzzyWatchHandle`) is deliberately deferred — proposed for `Subscribe`/`ListenConfig` jointly in #655.
3. `ListenConfig` on a closed client returns `ErrConfigClientClosed` instead of silently registering into a dead client.
4. The push handler acks unknown keys with a success response instead of returning nothing.

## Known inherited limitation

A server push that lands between an executor round's response and its sync-flag update can be overwritten and only recovered by the 3-minute full resync (the window is microseconds; the previous code had a strictly larger version of the same window across all task shards). Kept as-is, documented here.

## Testing

- TDD throughout; unit suites `-race -count=1` green, including targeted concurrency tests (listen/cancel hammer with invariant checks, cancel-during-drain, revive-during-cancel-in-flight, panic replay, watermark skip).
- Integration (`-tags=integration -race -count=1`) green against both Nacos 2.5.2 and 3.2.0, including the new multi-listener / cancel / CAS / server-restart tests.

### DIFF
--- a/clients/config_client/config_cache_holder.go
+++ b/clients/config_client/config_cache_holder.go
@@ -1,0 +1,163 @@
+/*
+ * Copyright 1999-2020 Alibaba Group Holding Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package config_client
+
+import (
+	"sync"
+
+	"github.com/nacos-group/nacos-sdk-go/v3/vo"
+)
+
+// cacheData holds the listen state for a single dataId/group/tenant tuple.
+// It is always referenced by pointer so that listeners registered against
+// the same key observe/update a single shared instance.
+type cacheData struct {
+	mu sync.Mutex
+
+	dataId, group, tenant                       string
+	content, contentType, md5, encryptedDataKey string
+	taskId                                      int
+	isSyncWithServer                            bool
+	isInitializing                              bool
+	discard                                     bool
+	listeners                                   []*listenerWrap
+}
+
+// listenerWrap pairs a user listener callback with the md5 watermark it has
+// last been notified with, so distinct listeners on the same key can be
+// notified independently.
+type listenerWrap struct {
+	listener    vo.Listener
+	lastCallMd5 string
+}
+
+// configCacheHolder is the pointer-based replacement for the previous
+// cache.ConcurrentMap-backed by-value storage. Entries are never replaced in
+// place; callers mutate the returned *cacheData under its own mu.
+type configCacheHolder struct {
+	mu      sync.RWMutex
+	entries map[string]*cacheData
+}
+
+func newConfigCacheHolder() *configCacheHolder {
+	return &configCacheHolder{
+		entries: make(map[string]*cacheData),
+	}
+}
+
+// get returns the entry for key, if any.
+func (h *configCacheHolder) get(key string) (*cacheData, bool) {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+	cData, ok := h.entries[key]
+	return cData, ok
+}
+
+// getOrCreate returns the existing entry for key, reviving it (discard=false)
+// if it was previously cancelled. If no entry exists yet, seed() is invoked
+// to build one, which is then stored and returned. seed is only invoked while
+// creating a brand-new entry.
+func (h *configCacheHolder) getOrCreate(key string, seed func() *cacheData) *cacheData {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	if cData, ok := h.entries[key]; ok {
+		cData.mu.Lock()
+		cData.discard = false
+		cData.mu.Unlock()
+		return cData
+	}
+	cData := seed()
+	h.entries[key] = cData
+	return cData
+}
+
+// snapshot returns all entries currently stored. Order is unspecified.
+func (h *configCacheHolder) snapshot() []*cacheData {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+	result := make([]*cacheData, 0, len(h.entries))
+	for _, cData := range h.entries {
+		result = append(result, cData)
+	}
+	return result
+}
+
+// removeIfDiscarded removes the entry for key only if it is still marked as
+// discarded and has no listeners attached, re-checking both conditions while
+// holding the lock to avoid racing with a concurrent revive/addListener.
+func (h *configCacheHolder) removeIfDiscarded(key string) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	cData, ok := h.entries[key]
+	if !ok {
+		return
+	}
+	cData.mu.Lock()
+	shouldRemove := cData.discard && len(cData.listeners) == 0
+	cData.mu.Unlock()
+	if shouldRemove {
+		delete(h.entries, key)
+	}
+}
+
+// count returns the number of entries currently stored.
+func (h *configCacheHolder) count() int {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+	return len(h.entries)
+}
+
+// addListener appends a listener to the entry, recording initialMd5 as its
+// starting watermark so it is only notified once content actually changes
+// relative to that baseline.
+func (c *cacheData) addListener(l vo.Listener, initialMd5 string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.listeners = append(c.listeners, &listenerWrap{listener: l, lastCallMd5: initialMd5})
+}
+
+// reviveAndAddListener atomically (re)activates the entry and appends l as a
+// new listener, in a single critical section: discard is reasserted to
+// false, isInitializing is set to true, and the listener is appended using
+// the entry's md5 (read under the same lock) as its initial watermark.
+//
+// This exists because getOrCreate's own revive step (discard=false) and a
+// subsequent, separately-locked addListener call are two distinct critical
+// sections; a concurrent CancelListenConfig/markDiscard could interleave
+// between them and leave the entry with discard==true and a non-empty
+// listeners slice -- an inconsistent state that downstream reap logic must
+// never observe. Folding revive + append into one lock closes that window:
+// whichever of markDiscard/reviveAndAddListener runs last atomically
+// determines the resulting (discard, listeners) pair.
+func (c *cacheData) reviveAndAddListener(l vo.Listener) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.discard = false
+	c.isInitializing = true
+	c.listeners = append(c.listeners, &listenerWrap{listener: l, lastCallMd5: c.md5})
+}
+
+// markDiscard flags the entry as cancelled: it stops being treated as synced
+// and its listeners are dropped. The entry itself is left in the holder for
+// removeIfDiscarded (or a future revive via getOrCreate) to reconcile.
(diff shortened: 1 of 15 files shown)


Sunrisea:
这里建议合并前修复一个 P1 回归：不要在配置监听主循环中同步执行用户回调。

[`notifyListeners` 这里直接调用 `deliverAndAdvance`](https://github.com/nacos-group/nacos-sdk-go/blob/d04173784a9f930e781731ffb6804ba4648a1ed4/clients/config_client/config_cache_holder.go#L210-L211)，调用链是 `startInternal → executeConfigListen → refreshContentAndCheck → notifyListeners → listener`。同一个 ConfigClient 只有一个监听执行循环，所以影响不只是同一配置下的多个 listener：任意一个慢回调都会阻塞其他配置的查询、通知，以及后续取消订阅请求。

针对当前 head `d041737`，用 mock RPC 驱动真实 `startInternal` 循环做了定向测试（`-race`）：

- 让一轮响应返回 A、B 两个变更，A 回调等待 B 回调通知：B 的查询都不会开始，形成循环等待。
- A 阻塞期间调用 `CancelListenConfig(C)`，API 返回 nil，但服务端的 `listen=false` 请求一直无法发出；释放 A 后，B 通知和 C 的取消请求才继续执行。
- 同期模拟 40 次配置推送，推送均得到成功响应，但 `asyncNotifyListenConfig` 的待发送 goroutine 持续积压，测试中观察到 41 个。

在 base `93a9350` 上，旧实现用 goroutine 执行回调，同样的“A 等待 B 通知”用例可以通过，不需要外部释放 A。因此这是已有 Go 行为的回归，而不只是同步/异步的风格选择。Java 默认回调可以同步执行，但它提供了 `Executor` 选择，不能据此直接改变 Go 的回调执行语义。

建议保留异步回调，并用每个 listener 的 in-flight 状态避免重复并发投递，成功后再推进 `lastCallMd5`。请补充覆盖真实监听循环的阻塞回调回归测试，验证一个配置的回调不会阻塞其他配置的处理。

cxhello:
Verified and fixed — this was a real regression, not a style choice. The base implementation launched every callback with `go listener(...)`, while this branch ran them inline on the single listen executor goroutine, so one slow callback stalled every other config's queries, notifications and cancel batches, and each pending bell parked a goroutine. Both of your test scenarios reproduced exactly as described.

Fix (commit 9dc559b), along the lines you suggested:

- **Callbacks run on their own delivery goroutine, never on the executor.** `notifyListeners` snapshots the entry under its lock, then launches one goroutine per lagging listener.
- **Per-listener `inFlight` gate.** A wrap whose previous delivery has not returned is skipped by subsequent rounds, so a single listener is never invoked concurrently with itself and a slow listener holds exactly one goroutine no matter how many changes arrive meanwhile. `lastCallMd5` advances only when the callback returns normally, in the same critical section that clears `inFlight`.
- **Catch-up path.** When a delivery completes and the wrap still trails the entry's current md5 (a newer change landed while the callback ran, or it panicked), the delivery rings the bell and the executor's round-entry sweep re-notifies — so a lagging listener catches up promptly instead of waiting out the poll interval.
- **Coalescing bell.** `listenExecute` is now a 1-buffered channel with a non-blocking send (same shape as the naming holder's `Bell`), which removes the goroutine-per-notification pile-up you observed entirely — your 40-push scenario now allocates zero goroutines for pending bells.

Regression tests drive the real `startInternal` loop against a scripted proxy (`-race`, ×5):

- `TestBlockingCallbackDoesNotBlockOtherConfigs` — A's blocked callback; B's change in a later round is still delivered while A is blocked (fails on the previous head, loop stalled).
- `TestCancelProceedsWhileCallbackBlocked` — `CancelListenConfig` gets its `listen=false` batch sent while A is blocked (fails on the previous head).
- `TestBellStormDoesNotAccumulateGoroutines` — 40 notifications with no executor running: goroutine count stays flat (fails on the previous head with +40).
- `TestNoConcurrentDuplicateDeliveryPerListener` and `TestLaggingWatermarkRedeliveredAfterInFlightCompletes` pin the in-flight gate and the catch-up semantics.

Full unit suite and both-version integration (2.5.2 / 3.2.0, `-race -count=1`) are green.

Sunrisea:
异步回调这部分修复已复核：相关回归测试本地 `-race -count=5` 通过，上次指出的跨配置阻塞问题已解决。

还有一个 P2 建议一并修复：持续 panic 的 listener 会触发无间隔重试。

[`deliverAndAdvance` 的这段逻辑](https://github.com/nacos-group/nacos-sdk-go/blob/9dc559b0da9dad74caa5d632970704ca60e28b37/clients/config_client/config_cache_holder.go#L249-L258)在回调 panic 后不推进 `lastCallMd5`，但仍因 `lagging` 立即调用 `wake()`；下一轮入口扫描又投递同一内容，形成“panic → wake → 再投递 → panic”的忙循环。容量为 1 的 bell 只能合并待处理通知，无法打断这种逐轮自唤醒。

本地用真实 `startInternal` 循环做了有界复现（`-race`）：缓存已与服务端同步，没有新推送，listener 前 99 次调用均 panic，第 100 次正常返回以终止探针，约 5ms 就完成了 100 次调用，并反复输出异常堆栈。持续 panic 时这条路径会一直消耗 CPU、刷错误日志。

建议采用最小修改：保留异步投递、`inFlight` 和每轮补发检查，只收紧主动唤醒条件：

```go
// 在 c.mu 内计算，在锁外调用 wake。
shouldWake := delivered && md5 != c.md5 && !c.discard
```

- 成功且执行期间出现了新版本：推进本次水位，立即唤醒以追上最新内容。
- panic：保留旧水位、清除 `inFlight`，不因失败主动唤醒，交给正常监听轮次重试。
- 过滤器失败：保持现有的保留水位、下一轮重试行为。

这与 Java SDK 的失败处理思路一致：[CacheData 捕获异常后只清理执行状态，不立即唤醒](https://github.com/alibaba/nacos/blob/b4fec02964c8606ad6b2ea954bd90d96023d4d00/client/src/main/java/com/alibaba/nacos/client/config/impl/CacheData.java#L577-L599)，后续由 [ClientWorker 的监听循环](https://github.com/alibaba/nacos/blob/b4fec02964c8606ad6b2ea954bd90d96023d4d00/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java#L1036-L1052)再次检查。没有其他事件时通常等约 5 秒；其他推送仍可能提前触发重试，这不是严格的失败冷却时间，也无需为此新增复杂的重试队列或退避机制。

建议补两个真实监听循环测试：持续 panic 且没有新事件时不会高速自重试；失败回调恢复后，即使服务端配置未再变化，也能在后续轮询中补发成功。现有“成功回调完成后追上新版本”的测试继续保留。

cxhello:
Confirmed and fixed in 48f3294. Reproduced first: the new busy-loop regression test (real `startInternal` loop, synced entry, persistently panicking listener, one initial bell) recorded **12,093 callback invocations in 600ms** on the previous head — the panic→wake→redeliver spin exactly as you described.

The fix is your minimal change, applied verbatim:

```go
shouldWake := delivered && md5 != c.md5 && !c.discard
```

computed inside `c.mu`, `wake()` invoked outside the lock. A successful delivery that fell behind a concurrent update still wakes immediately (the round-1 catch-up semantics are unchanged — `TestLaggingWatermarkRedeliveredAfterInFlightCompletes` still passes); a panicking delivery keeps its watermark, clears `inFlight`, and is retried only by the regular poll round's re-notify sweep; the filter-error path is untouched. Comments on `deliverAndAdvance`/`notifyListeners`/the executor sweep now state the failed-delivery-rides-the-poll contract with the Java `CacheData`/`ClientWorker` parity rationale.

Both requested tests drive the real `startInternal` loop against a scripted proxy (`-race`, ×5):

- `TestPersistentlyPanickingListenerDoesNotBusyLoop` — synced entry, no new events: a panicking listener gets at most the initial delivery within a 600ms window (12,093 → 1).
- `TestRecoveredListenerIsRedeliveredByPollWithoutServerChange` — the listener panics once and then recovers: with no further server change and no further bell, the poll round redelivers successfully and the watermark advances.

Full unit suite and both-version integration (2.5.2 / 3.2.0, `-race -count=1`) are green.

Issue #910, opened 2026-08-21
After server token secret key rotation, Go client never re-logins and keeps getting 403 Invalid signature


zhangxiaowei1225:
## Environment

- nacos-sdk-go version: v2.3.0 / v2.3.5 (same logic in latest master)
- Nacos server version: v2.5.1 (3-node cluster, auth enabled)
- Go: 1.22

## Description

After rotating `nacos.core.auth.plugin.nacos.token.secret.key` on the server
(rolling restart of a 3-node cluster), all previously issued access tokens
become invalid immediately.

- **Java SDK clients recovered within ~60s**: they receive 403 →
  `reLogin()` → new token.
- **Go SDK clients do NOT recover**: they keep sending
  `ConfigBatchListenRequest` with the stale token for hours, getting
  `403 Invalid signature` continuously, until the token-TTL time window
  passes or the process is restarted.

## Root cause

1. `common/security/nacos_auth_client.go` `login()` only checks the local
   refresh time window and thinks the token is still valid after a key
   rotation (default `tokenTtl=18000s`, so ~4h skip window):

   ```go
   tokenRefreshTime := ac.lastRefreshTime + ac.tokenTtl - 2*ac.tokenRefreshWindow
   if time.Now().Unix() < tokenRefreshTime {
       return true, nil
   }
   ```

2. `common/remote/rpc/rpc_client.go` `Request()` only logs a warning on a
   failure response (403) and never triggers re-login. The Java SDK does the
   opposite: `ClientWorker.requestProxy()` calls `reLogin()` on `NO_RIGHT` (403).

## Expected behavior

When the server responds 403 (`Invalid signature` / token invalid), the Go SDK
should trigger re-login immediately (with a rate-limit window to avoid login
storms), same as the Java SDK.

## Steps to reproduce

1. Start a Go config-listener client with username/password
   (e.g. nacos-sdk-go v2.3.0 or latest).
2. Rotate `nacos.core.auth.plugin.nacos.token.secret.key` on the server and
   rolling-restart the cluster.
3. Observe: the client keeps logging
   `ConfigBatchListenRequest ... error code: 403 ... [Invalid signature]`
   and never re-logins until the TTL window passes or the process restarts.

## Logs

```
2026-08-21 19:11:54,731 DEBUG access denied, request: ConfigBatchListenRequest, reason: Invalid signature
```

## Workaround

Restart the client process (it re-logins on startup).

PR #915, opened 2026-09-01
fix: re-login on 403 to recover from server token secret key rotation (#910)

zmz789:
**Related issue: #910**

## What is the purpose of the change

As reported in #910, after rotating `nacos.core.auth.plugin.nacos.token.secret.key` on the server, Go SDK clients keep sending requests with the stale token and receive `403 Invalid signature` until the token TTL window passes (~4h with the default `tokenTtl=18000s`) or the process restarts. Java SDK clients recover by re-login on 403.

This PR attempts to close that gap by following the Java SDK's design as closely as the current Go code structure allows. Since this touches the request hot path and the auth flow, I would appreciate careful review — especially on the behavior changes listed below.

## Root cause

1. `NacosAuthClient.login()` only checks the local refresh time window and has no way to learn that the server has rejected the token.
2. `RpcClient.Request()` never reacts to 403: an `ErrorResponse(403)` is swallowed by the retry loop (retries carrying the same stale token), and other failed responses only produce a warn log.

## Brief changelog

- `common/remote/rpc/rpc_client.go`: on `ErrorResponse` with `errorCode == 403`, call `nacosServer.ReLogin()` and return the response directly instead of entering the retry loop.
- `common/security/nacos_auth_client.go`: add `reLoginFlag` and `ReLogin()`. While the flag is set, `login()` bypasses the tokenTtl window and only respects a 60s debounce window (`RE_LOGIN_WINDOW_SECONDS`). A successful login clears the flag.
- `common/security/security_proxy.go`, `common/nacos_server/nacos_server.go`: add `ReLogin()` pass-through. `RamAuthClient` holds no token and is skipped via interface assertion.
- `common/constant/const.go`: add `NO_RIGHT = 403` (same name/value as Java `NacosException.NO_RIGHT`) and `RE_LOGIN_WINDOW_SECONDS = 60` (Java `reLoginWindow = 60000ms`).

## Points I am less sure about — review appreciated

1. **403 no longer retries / switches server.** Previously a 403 went through `waitReconnect` ×3 and then marked the client `UNHEALTHY`, triggering `switchServerAsync`. Retrying with the same stale token cannot succeed, and in testing this produced a reconnect loop (~3 reconnects/sec against a single server). Returning directly avoids that, but it *is* a behavior change: callers now see the 403 response sooner. If any downstream logic depends on the retry/switch behavior for 403, please flag it.
2. **`login success` log line moved.** The identical log previously lived in `NacosAuthClient.AutoRefresh`, which appears to be dead code (the live path is `SecurityProxy.AutoRefresh` → `Login()`). I moved it into `login()` so every real login is observable, and removed it from `AutoRefresh` to avoid double logging. Happy to revert this part if it is considered out of scope.
3. **60s debounce window semantics.** Aligned with Java (`reLoginWindow`, measured from `lastRefreshTime`). Consequence: if a 403 arrives within the first 60s after a login, recovery takes up to ~60s instead of ~5s. I kept Java's semantics intentionally, but I can make the window configurable or shorten it if preferred.
4. **Coverage is config-path focused.** The end-to-end verification below exercises the config listen path (`ConfigBatchListenRequest` / `ConfigQueryRequest`). The naming gRPC path shares the same `RpcClient.Request()` and `SecurityProxy`, so it should benefit identically, but I have not run a naming-specific end-to-end test.
5. **HTTP fallback path not covered.** The 403 handling here is at the gRPC layer. The HTTP fallback (`ReqConfigApi` in `nacos_server.go`) shares the same token source via `SecurityProxy`, so a triggered re-login also refreshes it, but I did not add explicit 403 handling there.

### Unit tests

- `TestNacosAuthClient_ReLogin`: login is a no-op inside the ttl window → after `ReLogin()` it performs a real login → flag cleared on success → repeated `ReLogin()` within 60s is throttled.
- `TestSecurityProxy_ReLogin`: flag propagated; RAM client skipped without panic.
- `TestRequest_NoRightReturnsDirectly`: a 403 `ErrorResponse` is returned directly with exactly 1 connection call (on unfixed code this test fails: 3 retries, then an error).

### End-to-end reproduction

To verify without rotating keys on a real server, I used a mock Nacos server (HTTP login issuing incrementing tokens + gRPC serving config requests, accepting only the latest issued token) behind a small reverse proxy that rewrites the `accessToken` in the first login response — this reproduces the client-side state after a key rotation (local TTL window still valid, server rejects the token).

Before this fix: `Send request fail ... error=Invalid signature` retried 3 times per listen round, followed by a reconnect loop, indefinitely; no re-login ever happened.

After this fix:

```
INFO  login success, tokenTtl: 18000 seconds, ...              (login #1)
WARN  ConfigQueryRequest request rejected by server, error code: 403, message: [Invalid signature], mark re-login
WARN  ConfigBatchListenRequest request rejected by server, ... (every 5s, within the 60s debounce window)
INFO  login success, tokenTtl: 18000 seconds, ...              (login #2, ~60s after the first 403)
(server side) ConfigBatchListenRequest token="real-token-2" -> OK   (recovered)
```

## Compatibility

- Only additive methods (`ReLogin()`); no existing signatures changed.
- No behavior change when auth is not configured (no username): `login()` returns early as before.
- `RamAuthClient` is unaffected: it computes per-request signatures and holds no token.

### DIFF
--- a/common/constant/const.go
+++ b/common/constant/const.go
@@ -93,6 +93,8 @@ const (
 	LABEL_MODULE_NAMING               = "naming"
 	RESPONSE_CODE_SUCCESS             = 200
 	UN_REGISTER                       = 301
+	NO_RIGHT                          = 403
+	RE_LOGIN_WINDOW_SECONDS           = 60
 	KEEP_ALIVE_TIME                   = 5
 	DEFAULT_TIMEOUT_MILLS             = 3000
 	ALL_SYNC_INTERNAL                 = 5 * time.Minute
--- a/common/nacos_server/nacos_server.go
+++ b/common/nacos_server/nacos_server.go
@@ -348,6 +348,10 @@ func (server *NacosServer) InjectSecurityInfo(param map[string]string, resource 
 	}
 }
 
+func (server *NacosServer) ReLogin() {
+	server.securityLogin.ReLogin()
+}
+
 func getAddress(cfg constant.ServerConfig) string {
 	if strings.Index(cfg.IpAddr, "http://") >= 0 || strings.Index(cfg.IpAddr, "https://") >= 0 {
 		return cfg.IpAddr + ":" + strconv.Itoa(int(cfg.Port))
--- a/common/remote/rpc/rpc_client.go
+++ b/common/remote/rpc/rpc_client.go
@@ -624,6 +624,14 @@ func (r *RpcClient) Request(request rpc_request.IRequest, timeoutMills int64) (r
 			currentErr = waitReconnect(timeoutMills, &retryTimes, request, err)
 			continue
 		}
+		if response != nil && (response.GetErrorCode() == constant.NO_RIGHT || response.GetResultCode() == constant.NO_RIGHT) {
+			if r.nacosServer != nil {
+				r.nacosServer.ReLogin()
+			}
+			r.lastActiveTimestamp.Store(time.Now())
+			return response, errors.Errorf("%s request rejected by server, error code: %d, result code: %d, message: [%s]",
+				request.GetRequestType(), response.GetErrorCode(), response.GetResultCode(), response.GetMessage())
+		}
 		if resp, ok := response.(*rpc_response.ErrorResponse); ok {
 			if resp.GetErrorCode() == constant.UN_REGISTER {
 				r.mux.Lock()
@@ -634,23 +642,10 @@ func (r *RpcClient) Request(request rpc_request.IRequest, timeoutMills int64) (r
 				}
 				r.mux.Unlock()
 			}
-			if resp.GetErrorCode() == constant.NO_RIGHT {
-				logger.Warnf("%s request rejected by server, error code: %d, message: [%s], mark re-login",
-					request.GetRequestType(), resp.GetErrorCode(), resp.GetMessage())
-				if r.nacosServer != nil {
-					r.nacosServer.ReLogin()
-				}
-				return response, nil
-			}
 			currentErr = waitReconnect(timeoutMills, &retryTimes, request, errors.New(response.GetMessage()))
 			continue
 		}
 		if response != nil && !response.IsSuccess() {
-			if response.GetErrorCode() == constant.NO_RIGHT || response.GetResultCode() == constant.NO_RIGHT {
-				if r.nacosServer != nil {
-					r.nacosServer.ReLogin()
-				}
-			}
 			logger.Warnf("%s request received fail response, error code: %d, result code: %d, message: [%s]",
 				request.GetRequestType(), response.GetErrorCode(), response.GetResultCode(), response.GetMessage())
 		}
--- a/common/remote/rpc/rpc_client.go
+++ b/common/remote/rpc/rpc_client.go
@@ -634,11 +634,25 @@ func (r *RpcClient) Request(request rpc_request.IRequest, timeoutMills int64) (r
 				}
 				r.mux.Unlock()
 			}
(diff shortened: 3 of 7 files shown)


Sunrisea:
Thanks for working on this. I compared the implementation with the current Java SDK and the Nacos 2.5.1 server behavior.

The overall re-login design looks aligned with Java: mark the authentication state on 403, let the periodic login task refresh the token, and use a 60-second window to avoid login storms.

There is one blocking issue: **the current request must still be reported as failed after marking re-login.**

Nacos 2.5.1 normally returns a handler-specific response with `errorCode = 403`, rather than always returning a generic `ErrorResponse`:

- [Nacos 2.5.1 auth filter](https://github.com/alibaba/nacos/blob/3d2f10b2398411797b6b1501f811dc84bfbcbd2f/core/src/main/java/com/alibaba/nacos/core/auth/RemoteRequestAuthFilter.java#L97-L104)
- The Java Config client marks re-login and lets the caller handle the failed typed response: [ClientWorker](https://github.com/alibaba/nacos/blob/915ff1fb243fe9992fbb94233638013c587c66e1/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java#L1390-L1400)
- The Java Naming client marks re-login and throws before converting the response to a concrete response type: [NamingGrpcClientProxy](https://github.com/alibaba/nacos/blob/915ff1fb243fe9992fbb94233638013c587c66e1/client/src/main/java/com/alibaba/nacos/client/naming/remote/gprc/NamingGrpcClientProxy.java#L525-L540)

The current Go implementation returns a 403 response with a `nil` error:

```go
return response, nil
```

This causes two problems:

1. For a generic `ErrorResponse`, Naming methods that directly assert a concrete response type may panic.
2. For the production-shaped typed 403 response, Naming methods may report a rejected request as `false, nil`, empty data with `nil`, or even `nil` from `Unsubscribe`.

I suggest keeping the intended “no retry and no server switch” behavior, but returning a non-nil error after marking re-login:

```go
if response != nil &&
	(response.GetErrorCode() == constant.NO_RIGHT ||
		response.GetResultCode() == constant.NO_RIGHT) {
	if r.nacosServer != nil {
		r.nacosServer.ReLogin()
	}
	r.lastActiveTimestamp.Store(time.Now())
	return response, errors.Errorf(
		"%s request rejected by server: error code=%d, result code=%d, message=%s",
		request.GetRequestType(),
		response.GetErrorCode(),
		response.GetResultCode(),
		response.GetMessage(),
	)
}
```

This preserves the Java SDK semantics:

- mark re-login;
- do not retry with the stale token;
- keep the RPC connection healthy;
- report the current operation as failed;
- only process or convert successful responses.

Please also cover both response shapes in tests:

- a typed response with `errorCode = 403`, matching the Nacos 2.5.1 auth-filter path;
- a generic `ErrorResponse` with `errorCode = 403`.

The tests should verify one underlying request, a non-nil error, the client remaining `RUNNING`, and no Naming panic or silent success.

No synchronous login is needed in the RPC request path; the existing background refresh flow can perform the actual re-login.

zmz789:
Thanks for the careful review. I have pushed a fix that keeps the "no retry, no server switch" behavior but fails the current request explicitly. Both response shapes are covered by tests.

For the "only process or convert successful responses" point, I audited all callers of `Request()` from naming side and config side, confirmed a type-assertion panic or silent processing of a failed response will not occur.

Sunrisea:
lgtm

